PR #186, opened 2019-04-22
✅ Add test renderer package

dankolbman:
Adds the `react-test-renderer` package so that we may write and run snapshot tests.


dankolbman:
Superseded by #186.

PR #203, opened 2019-04-24
🤡 Mock UI Kit Icon component

dankolbman:
This mocks out the `Icon` component from the UI Kit so that the full svg will not be rendered to the snapshot thus creating a much neater snapshot.
Also updates snapshots for the mock and for changes introduced in #201


dankolbman:
@bdolly we may also be able to roll out mocks with the library itself.

PR #214, opened 2019-04-27
👷 Add bundlesize status check

dankolbman:
Adds bundlesize CI status check to keep our bundle size in check.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,16 @@
   "name": "kf-ui-data-tracker",
   "version": "0.2.0",
   "private": true,
+  "bundlesize": [
+    {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -2,6 +2,16 @@
   "name": "kf-ui-data-tracker",
   "version": "0.1.0",
   "private": true,
+  "bundlesize": [
+    {
+      "path": "./build/static/js/*.js",
+      "maxSize": "3 MB"
```
bdolly: are these just arbitrary values?

dankolbman: yeah, the bundlesize check needs them.

Issue #155, opened 2019-04-09
Badges look like buttons


allisonheath:
![image](https://user-images.githubusercontent.com/2213854/55813894-8ea3e280-5abb-11e9-83ed-cb2f66c162ab.png)

dankolbman:
Should consider this before proceeding with kids-first/kf-uikit#107
@XuTheBunny @bdolly

bdolly:
to my knowledge these aren't functional yet, as in they are just hard coded and there's nothing implemented to change their status. @dankolbman is that correct?

PR #218, opened 2019-04-30
♻️ Study File List improvements

### LINKED ISSUE #169: View description in file list
Screen Shot 2019-04-11 at 11 23 00 AM

Was sent this link from Bailey to look at one of the Maris manifests, I'd like to quickly view the descriptions inline so I can determine which one of the manifests I was interested in. In this case it was the one with the description:

>From Salina Hall @ HA on 4/5/19. List of flow cells and/or samples and/or libraries that we should not analyze because there is something wrong with the data for one reason or another. 

versus

>From Salina Hall @ HA on 4/5/19. List of samples that were not sequenced. We should still double check if we received data for them. 

Currently, I have to view each description in the separate edit window (and only because I've uploaded files already myself did I realize this was the right place to go).

bdolly:
Restructures the Single Study File List and adds truncated file descriptions to list items

Closes #169

### DIFF
--- a/src/common/fileUtils.js
+++ b/src/common/fileUtils.js
@@ -1,0 +1,21 @@
+// TODO: add doc blocks
+
+export const dateCompare = (version1, version2) => {
+  return version1.node.createdAt < version2.node.createdAt;
+};
+
+export const formatFileSize = (bytes, si) => {
+  var thresh = si ? 1000 : 1024;
+  if (Math.abs(bytes) < thresh) {
+    return bytes + ' B';
+  }
+  var units = si
+    ? ['kB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB']
+    : ['KiB', 'MiB', 'GiB', 'TiB', 'PiB', 'EiB', 'ZiB', 'YiB'];
+  var u = -1;
+  do {
+    bytes /= thresh;
+    ++u;
+  } while (Math.abs(bytes) >= thresh && u < units.length - 1);
+  return bytes.toFixed(1) + ' ' + units[u];
+};
--- a/src/components/FileList/FileElement.js
+++ b/src/components/FileList/FileElement.js
@@ -2,32 +2,15 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classes from 'classnames';
 import {withRouter, Link} from 'react-router-dom';
-import TimeAgo from 'react-timeago';
-import {Icon, GridContainer} from 'kf-uikit';
-import Badge from '../Badge/Badge';
 import CopyButton from '../CopyButton/CopyButton';
+import TimeAgo from 'react-timeago';
+import Badge from '../Badge/Badge';
+import {GridContainer} from 'kf-uikit';
+import FileActionsContainer from '../../containers/FileActionsContainer';
+import {dateCompare, formatFileSize} from '../../common/fileUtils';
 /**
- * Displays unordered studies in grid view (include empty stage message)
+ * Displays unordered study files in list view
  */
-const dateCompare = (version1, version2) => {
-  return new Date(version2.node.createdAt) - new Date(version1.node.createdAt);
-};
-
-const formatFileSize = (bytes, si) => {
-  var thresh = si ? 1000 : 1024;
-  if (Math.abs(bytes) < thresh) {
-    return bytes + ' B';
-  }
-  var units = si
-    ? ['kB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB']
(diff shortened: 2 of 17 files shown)


dankolbman:
I'm not sure if this addresses #155? @allisonheath Do the badges still look like buttons to you?

src/common/fileUtils.js
```diff
@@ -0,0 +1,21 @@
+// TODO: add doc blocks
+
+export const dateCompare = (version1, version2) => {
+  return version1.node.createdAt < version2.node.createdAt;
```
dankolbman: Make sure this get updated from the change in #216

src/components/FileList/FileElement.js
```diff
@@ -62,15 +36,19 @@ const FileElement = ({
   const fileNameClass = classes(
     {
       'FileStatus--loading': loading,
-      FileStatus: !loading,
+      'FileList--Element-name': !loading,
```
dankolbman: Would `name` be the element in the BEM format?

src/components/FileList/FileElement.js
```diff
@@ -86,57 +64,47 @@ const FileElement = ({
...
+        </div>
+
+        <div className="cell-9 h-full">
+          <Link
+            to={`/study/${match.params.kfId}/files/${fileKfID}`}
+            className="w-full no-underline text-black inline-block"
+          >
+            <p className={fileNameClass}>{fileName}</p>
+            {fileDescription ? (
+              <p className="FileList--Element-desc w-3/4 h-8 m-0 p-0 text-truncate text-black">
```
dankolbman: `FileList--Description`?

src/components/FileList/filelist.css
```diff
@@ -1,11 +1,39 @@
...
-  @inherit: .list-reset, .min-w-full, .my-20, .border-t, .border-grey;
+  @inherit: .list-reset, .min-w-full, .my-20, .p-0, .ml-0, .border-t,
+    .border-grey;
 }
 
 .FileList--Element {
-  @inherit: .list-reset, .w-full, .border-b, .border-grey, .py-8, .px-32, .flex,
-    .flex-row, .flex-wrap-reverse;
+  @inherit: .list-reset, .w-full, .py-8, .border-b, .border-grey, .text-black;
+  transition: all 0.25s ease-in;
```
dankolbman: The animation is pretty frustrating and having the description hidden makes it difficult to scan the files for the one you one and compare across them.
Are there other approaches for showing the description that allows that?

src/components/FileList/FileElement.js
bdolly: that would mean it's describing the `FileList` whereas this really describes the `FileList--Item` or `Element`

src/components/FileList/FileElement.js
bdolly: not sure, we really need to nail down our naming conventions and document them in the kf-uikit

src/common/fileUtils.js
bdolly: > Make sure this get updated from the change in #216

is this a general comment of the PR or specifically for the `dateCompare` function

src/components/FileList/filelist.css
bdolly: sped the animation up. If/when we get pagination in place we could potentially have the truncated desc without the hover state. As of now I think that showing the file desc is adding too much to clutter to a simple list item

src/common/fileUtils.js
dankolbman: Just making sure it makes it through the rebase

allisonheath:
The badges do still look like buttons, but because things are more lined up, it's way more readable making it more obvious they're status. I think I'd still advocate for the file name (or potentially the file type) being the first text to the left and the status being somewhere else in the table.

bdolly:
> The badges do still look like buttons, but because things are more lined up, it's way more readable making it more obvious they're status. I think I'd still advocate for the file name (or potentially the file type) being the first text to the left and the status being somewhere else in the table.

I think the badges look like buttons because we have a _very_ basic button style of text on a flat colored pill. The badges we're intentionally put on the left as it was _assumed_ that would be the most important piece of information for our primary user, the investigator. As of now there is no research or testing to validate this so I worry that we are making changes for our selves and not our users.

allisonheath:
> > The badges do still look like buttons, but because things are more lined up, it's way more readable making it more obvious they're status. I think I'd still advocate for the file name (or potentially the file type) being the first text to the left and the status being somewhere else in the table.
> 
> I think the badges look like buttons because we have a _very_ basic button style of text on a flat colored pill. The badges were intentionally put on the left as it was _assumed_ that would be the most important piece of information for our primary user, the investigator. As of now, there is no research or testing to validate this so I worry that we are making changes for our selves and not our users.

Who made that assumption in the first place? As you say, not really validated, so what's the difference in making the change at this point?

PR #221, opened 2019-05-01
♻️ Refactor copy button

### LINKED ISSUE #200: add icon for Copy-to-clipboard instead of checkmark character
It would be great if we could polish the `copied` state of this a bit to more closely match the github style.
 

By using an icon and maybe a tooltip we can remove the alert box. While no confirmation of checkmark icon exist in the design system this could be a great place to propose one.

XuTheBunny:
Replace the old checkmark with green icon svg.
Replace the "Copied!" alert with tooltip underneath the checkmark.

On study header:
image

On file list:
image

Closes #200

### DIFF
--- a/src/components/CopyButton/CopyButton.test.js
+++ b/src/components/CopyButton/CopyButton.test.js
@@ -29,19 +29,19 @@ it('renders correctly', () => {
   expect(tree).toMatchSnapshot();
 });
 
-it('displays ✔ on click', () => {
+it('displays checkmark svg on click', () => {
   jest.useFakeTimers();
   const tree = render(<CopyButton text="testing" />);
 
   // There should be no ✔ yet, only an Icon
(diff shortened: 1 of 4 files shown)


src/components/CopyButton/CopyButton.test.js
```diff
@@ -34,21 +34,21 @@ it('displays ✔ on click', () => {
...
 
   // There should be no ✔ yet, only an Icon
-  expect(tree.queryByText('✔')).toBeNull();
+  expect(tree.container.querySelector('span.CopyButton')).toBeNull();
   expect(tree.container.firstChild).toMatchSnapshot();
 
   let button = tree.container.querySelector('button');
   button.click();
 
   // Icon should be replaced with ✔
```
dankolbman: Can you update this comment now that the behavior has been changed and there is no longer a ✔ character?

PR #231, opened 2019-05-08
✨ Add pagination on file list 

### LINKED ISSUE #104: Paginate file listing
Presumably, the amount of files in some studies will grow to be very large. We should add features like pagination to help reduce the amount of displayed information on a single page.

XuTheBunny:
image
image

Closes #104

### DIFF
--- a/src/components/FileList/FileList.js
+++ b/src/components/FileList/FileList.js
@@ -1,24 +1,58 @@
-import React from 'react';
+import React, {useState, Fragment} from 'react';
 import PropTypes from 'prop-types';
 import classes from 'classnames';
 import FileElement from './FileElement';
-
+import ReactPaginate from 'react-paginate';
 /**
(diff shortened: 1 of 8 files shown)


bdolly:
This is looking good. I will provide some style direction for this later this week.

For now can we add a showing `X of X files` to it so the user knows the total number of files?

src/components/FileList/FileList.js
```diff
@@ -1,82 +1,113 @@
...
-              >
-                {(deleteFile, {loading, error}) => (
-                  <FileElement
-                    key={node.kfId}
-                    fileNode={node}
-                    loading={loading}
-                    error={error}
-                    deleteFile={() =>
-                      deleteFile({variables: {kfId: node.kfId}})
+    <div>
```
dankolbman: Could you use a Fragment here?

PR #237, opened 2019-05-15
✨ Add modal view for version upload and annotation

### LINKED ISSUE #177: Allow upload of new versions
The user should be able to upload a new version of a file to have it become the most recent version of the file. This should be done via a button on the file's page (#47)

### LINKED ISSUE #241: selecting any of the file type boxes causes all preceding options to shift position slightly
Safari browser

Screen-Recording-2019-05-16-at-12 39 46-PM

### LINKED ISSUE #243: Move the svgs to their own file
move the svgs to their own file
Ref: https://facebook.github.io/create-react-app/docs/adding-images-fonts-and-files#adding-svgs

XuTheBunny:
image

image

Closes #177
Closes #243 
Closes #241

### DIFF
--- a/src/components/FileAnnotation/FileEditor.js
+++ b/src/components/FileAnnotation/FileEditor.js
@@ -37,7 +37,7 @@ const FileEditor = ({
         {editingName ? (
           <label
             className="row-1 cell-12 flex justify-between"
-            for="file_name_input"
+            htmlFor="file_name_input"
           >
             <input
               data-testid="name-input"
--- a/src/components/FileDetail/AnnotateVersionModalContainer.js
+++ b/src/components/FileDetail/AnnotateVersionModalContainer.js
@@ -1,0 +1,51 @@
+import React from 'react';
+import FileEditorContainer from '../../containers/FileEditorContainer';
+import FileEditor from '../FileAnnotation/FileEditor';
+import Modal from '../Modal/Modal';
+
+const AnnotateVersionModalContainer = ({
+  kfId,
+  additionalContent,
+  onCloseDialog,
+}) => {
+  return (
+    <FileEditorContainer kfId={kfId}>
+      {({
+        updateFile,
+        fileByKfId,
+        fileNameInput,
+        fileType,
+        selectFileType,
+        onSubmit,
+        setFileName,
+        setFileDescription,
+      }) => (
+        <Modal
+          title="EDIT FILE INFO"
+          onSubmit={e => onSubmit(e, updateFile, onCloseDialog)}
+          onCloseModal={onCloseDialog}
+        >
+          <p className="font-title mt-0">
+            Help ensure the fastest processing and harmonization of your study
+            file by telling us about its contents. This helps our engineers
+            accurately interpret your data.
+          </p>
+          <div className="FileDialog--Edit">
+            <FileEditor
+              kfId={fileByKfId.kfId}
+              name={fileNameInput}
+              description={fileByKfId.description}
+              fileType={fileType}
+              selectFileType={selectFileType}
+              onNameChange={e => setFileName(e.target.value)}
+              onDescriptionChange={e => setFileDescription(e.target.value)}
+            />
+          </div>
+          {additionalContent && additionalContent()}
+        </Modal>
+      )}
+    </FileEditorContainer>
+  );
+};
+
+export default AnnotateVersionModalContainer;
--- a/src/components/FileDetail/FileDetailModal.js
+++ b/src/components/FileDetail/FileDetailModal.js
@@ -1,0 +1,77 @@
+import React, {Fragment} from 'react';
+import UploadVersionModalContainer from './UploadVersionModalContainer';
+import AnnotateVersionModalContainer from './AnnotateVersionModalContainer';
+import NotificationBar from './NotificationBar';
+import SvgIcon from '../Icon/Icon';
+/**
+ * Render either UploadVersionModal or AnnotateVersionModal
+ * Handling conditional rendering in a separate file to keep FileDetail clean
+ * Adding NotificationBar here since it is mainly used to switch modal view
+ */
+const NotificationAction = ({onClick, buttonText}) => (
+  <button
+    onClick={onClick}
+    className="FileVersionList--Button"
+    data-testid="notificationBar-button"
+  >
+    <SvgIcon kind="Edit" height="8" width="8" className="mr-4 mt-4" />
+    {buttonText}
+  </button>
+);
+
+const FileDetailModal = ({
+  className,
+  onSubmit,
+  children,
+  studyId,
+  fileNode,
+  dialog,
+  onCloseModal,
+  onNotificationClick,
+}) => {
+  return (
+    <Fragment>
+      {dialog === 'upload' && (
+        <UploadVersionModalContainer
+          studyId={studyId}
+          fileNode={fileNode}
+          onCloseDialog={onCloseModal}
+          additionalContent={() => (
+            <NotificationBar
+              mainButton={() => (
+                <NotificationAction
+                  onClick={onNotificationClick}
+                  buttonText="EDIT FILE INFO"
+                />
+              )}
+              icon="note"
+              message="Info Update?"
+              note="Want to change the Name, Description, or File Type as well?"
+            />
+          )}
+        />
+      )}
+      {dialog === 'annotation' && (
+        <AnnotateVersionModalContainer
+          kfId={fileNode.kfId}
+          onCloseDialog={onCloseModal}
+          additionalContent={() => (
+            <NotificationBar
+              mainButton={() => (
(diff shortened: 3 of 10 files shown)


dankolbman:
`💄 Minor style update the align header logo` Could go in its own PR since it's well encapsulated and addresses a specific issue not related to the modal feature.

src/components/FileDetail/NotificationBar.js
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classes from 'classnames';
+/**
+ * Displays notification bar with one main button (current)
+ */
+const NotificationBar = ({className, mainButton, icon, message, note}) => {
```
dankolbman: ✅ Is there testing for this?

src/components/FileDetail/AnnotateVersionModalContainer.js
```diff
@@ -0,0 +1,92 @@
+import React, {useState} from 'react';
+import {Query, Mutation} from 'react-apollo';
+import {UPDATE_FILE} from '../../state/mutations';
+import {GET_FILE_BY_ID} from '../../state/queries';
+import FileEditor from '../FileAnnotation/FileEditor';
+import {LoadingPlaceholder} from '../Loading';
+import Modal from '../Modal/Modal';
+
+const AnnotateVersionModalContainer = ({
```
dankolbman: This could be moved to `src/containers`

dankolbman: ✅ Is there testing for this?

src/components/FileDetail/UploadVersionModalContainer.js
```diff
@@ -0,0 +1,127 @@
+import React, {useState} from 'react';
+import {Query, Mutation} from 'react-apollo';
+import {GET_FILE_BY_ID} from '../../state/queries';
+import {CREATE_FILE} from '../../state/mutations';
+import {LoadingPlaceholder} from '../Loading';
+import FileUploadTarget from '../FileUpload/FileUploadTarget';
+import Modal from '../Modal/Modal';
+
+const UploadVersionModalContainer = ({
```
dankolbman: This could be moved to `src/containers`

dankolbman: ✅ Is there testing for this?

src/containers/UploadContainer.js
```diff
@@ -51,44 +57,21 @@ const UploadContainer = props => {
...
-          onDragEnter={handleDragEnter}
-          onDragLeave={handleDragLeave}
-          handleDrop={e => handleDrop(e, createFile)}
-          handleSelectedFile={e => handleSelectedFile(e, createFile)}
-          instructions="To upload files, drag and drop them here"
-        />
-      )}
+    <Mutation mutation={CREATE_FILE} refetchQueries={res => refetchAfterUpload}>
+      {(createFile, {data, error}) =>
+        children
```
dankolbman: Since you only need to render if there are children, you can use {children && (<div></div>)}` instead of the ternary.

src/components/FileAnnotation/FileEditor.js
```diff
@@ -79,55 +84,20 @@ const FileEditor = ({
...
-          <SelectElement
-            name="other"
-            value="OTH"
-            icon="settings"
-            title="Other"
-            body="Some helpful description."
-            select={selectFileType}
-            selected={fileType === 'OTH'}
-          />
+          {Object.keys(fileTypeDetail).map(item => (
```
dankolbman: Refactoring this should happen in a different pr.

src/components/FileDetail/FileDetailModal.js
```diff
@@ -0,0 +1,69 @@
...
+import {ReactComponent as Edit} from '../../assets/icons/edit.svg';
+
+const NotificationAction = ({onClick, buttonText}) => (
+  <button onClick={onClick} className="FileVersionList--Button">
+    <Edit className="mr-4" />
+    {buttonText}
+  </button>
+);
+
+const FileDetailModal = ({
```
dankolbman: ✅ Is there testing for this?

dankolbman: Could use some description about what this component does.

src/components/FileDetail/FileDetailModal.js
```diff
@@ -0,0 +1,69 @@
...
+  studyId,
+  fileNode,
+  dialog,
+  onCloseModal,
+  onNotificationClick,
+}) => {
+  return (
+    <Fragment>
+      {dialog === 'upload' && (
+        <UploadVersionModalContainer
```
dankolbman: What's the point of this component if it either renders a raw `UploadVersionModalContainer` or a raw `AnnotateVersionModalContainer`?

src/components/FileDetail/AnnotateVersionModalContainer.js
```diff
@@ -0,0 +1,51 @@
...
+              kfId={fileByKfId.kfId}
+              name={fileNameInput}
+              description={fileByKfId.description}
+              fileType={fileType}
+              selectFileType={selectFileType}
+              onNameChange={e => setFileName(e.target.value)}
+              onDescriptionChange={e => setFileDescription(e.target.value)}
+            />
+          </div>
+          {additionalContent && additionalContent()}
```
dankolbman: Why not use `children`?

src/containers/FileEditorContainer.js
```diff
@@ -49,28 +52,28 @@ const FileEditorContainer = ({kfId, history, match}) => {
...
-                  fileType={fileType}
-                  selectFileType={selectFileType}
-                  onSubmit={e => onSubmit(e, updateFile)}
-                  onNameChange={e => setFileName(e.target.value)}
-                  onDescriptionChange={e => setFileDescription(e.target.value)}
-                />
-              );
-            }}
+            {(updateFile, {_}) =>
+              children
```
dankolbman: Since you only want to render if there are children, you don't need the ternary and could just do `{children}`

src/containers/FileEditorContainer.js
```diff
@@ -49,28 +52,28 @@ const FileEditorContainer = ({kfId, history, match}) => {
...
+                ? children({
```
dankolbman: This is a bit confusing, typically `children` is any jsx element, but it looks like you are expecting it to be of a specific component and calling it functionally instead of using standard jsx syntax.
What was wrong with simply returning a `<FileEditor />` component?

src/views/__tests__/FileUpload.test.js
```diff
@@ -32,9 +32,10 @@ it('edits a file correctly', async () => {
   await wait();
 
   // Should be on the annotation page, edit the file name
-  fireEvent.click(tree.queryByTestId('edit-name-button'));
-
-  await wait();
+  // fireEvent.click(tree.queryByTestId('edit-name-button'));
```
dankolbman: Can you enable these?

dankolbman:
Looks like `♻️ Refactors file upload container `  could be squashed with `✨ Add file detail modal with annotation/upload container`

dankolbman:
`💄 Update file type images to match the design` Should be moved to a new PR.

dankolbman:
Any way that `✨ Add notification bar component` could be moved to a different PR?
There's a lot going on in this request, so it would be nice to narrow it down to only the fundamental changes needed for the modal.

src/components/FileDetail/FileDetailModal.js
bdolly: @dankolbman my thinking was too simple shorten the length of `FileDetail.js` by abstracting it out, but open to suggestions

src/components/FileDetail/UploadVersionModalContainer.js
bdolly: would really like to reorg the folder structure somehow. I feel like the container concept isn't scaling the best or maybe I'm going about it wrong

src/containers/FileEditorContainer.js
bdolly: we needed a way to add content before or after the editor in the context of the modal. Removing it here also helped to not duplicate the submission logic. 
When the FileEditor is inside of the modal the submission/save functionality is handled by the Modals footer buttons and not the form itself, so by removing the FileEditor here you can wrap the Modal and pass that logic down without needed the modal to handle the state for it.

Also open to another approach to this.

src/containers/FileEditorContainer.js
bdolly: if `children` is null it throws an error though

src/components/FileDetail/AnnotateVersionModalContainer.js
bdolly: `children` seeemd less semantic as to where the content would be rendered

Issue #244, opened 2019-05-20
🐛 postCss inherit causes ... inheritance problems 


bdolly:
## Current Behavior
The use of the [postcss-inherit](https://www.npmjs.com/package/postcss-inherit) to utilize uikit css declarations currently alters the output selector specificity caused collisions and inheritance headaches as the app scales. As you can see below since the **output.css** now has combined `.bg-black` and `.Card--Login` they have the same declaration set causing adverse side-effects when just using the `.bg-black` class which is undesired. 
```css
// uikit styles
.bg-black{
  background-color: #000;
}

// kf-ui-data-tracker styles 

--- input.css ---
.Card--Login{
  display: inline-block;
  overflow: hidden;
  width: 100%;
  @inherit: .bg-black;
...
}

--- output.css ---
.bg-black,
.Card--Login{
  display: inline-block;
  overflow: hidden;
  width: 100%;
  background-color: #000;
...
}
```

## Expected/Desired Behavior 
We would like to be able to use individual css declarations from the uikit, e.g.`color: #008299;`, to compose new css compoents inside the app. This allows us to maintain consistent declaration values across our styles while providing a high degree of flexibility. This behavior would simple add the uikit declaration to selector's declarations without altering the selector itself or its specificity
```css
// uikit styles
.bg-black{
  background-color: #000;
}

// kf-ui-data-tracker styles 

--- input.css ---
.Card--Login{
  display: inline-block;
  overflow: hidden;
  width: 100%;
  @expand .bg-black;
...
}

--- output.css ---
Card--Login{
  display: inline-block;
  overflow: hidden;
  width: 100%;
  background-color: #000;
...
}
```

PR #245, opened 2019-05-20
🚑 Fix postcss inheritance

bdolly:
💥 This introduces a breaking change in the way uikit styles are utilized 

Replaces [postcss-inherit](https://www.npmjs.com/package/postcss-inherit) with [postcss-extend-rule](https://github.com/jonathantneal/postcss-extend-rule) to solve css inheritance issues caused when using uikit delcarations (see #244 for details)


bdolly:
> Does this mean we can get rid of the `postcss-inherit` dependency in `package.json`?

Yeah I did `yarn remove postcss-inherit` which I thought re-wrote the package.json but maybe not

> Could you make sure the first commit only has one emoji in the summary?
Which emoji do you want? 
Yea

PR #249, opened 2019-05-22
💄 Align header to page content

### LINKED ISSUE #240: body position on coming soon panes doesn't match the files pane
comparison

XuTheBunny:
So the header logo can align with the main content

BEFORE:
comparison

AFTER:
image

Closes #240


dankolbman:
I think you want to reword to something like 'align navigation to page content'.

PR #258, opened 2019-05-24
✨ Add general svg icon rendering component

XuTheBunny:
Having all the svg icons saved as `.svg` file. 
Create a general component `<SvgIcon >` to render all different icons
Implement `<SvgIcon >` to all the existing component and update their snapshot test
Minor fix on uploadFile test by targetting document to render
Minor adjustment on version list style


src/views/__tests__/FileUpload.test.js
```diff
@@ -27,33 +45,33 @@ it('edits a file correctly', async () => {
   await wait();
 
   // Click on the file's edit button
```
bdolly: 🔥

bdolly:
looks like you can drop these commits in a rebase: 
- ecd7d8d
- 86aac80
- 672db5b
 rename this like so: 
- cc6b65f  📷 ->  ✅

Then just run `yarn test -u` and make a final commit that updates all snapshots at once

src/assets/icons/icon-box.svg
```diff
@@ -0,0 +1,12 @@
+<svg width="24px" height="24px" viewBox="0 0 24 24" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
bdolly: unfortunately, we have to go back through and add the  `Icon` class to all the **svg** elements and the add the `<iconName>-icon` id to the top level **g**(group tag) like how we have it in the uikit. https://github.com/kids-first/kf-uikit/blob/master/src/components/Icon/Icon.jsx

Otherwise we won't be able to uniformly style these like the other icons

src/components/FileAnnotation/SelectElement.js
```diff
@@ -31,10 +28,12 @@ const SelectElement = ({className, value, select, selected}) => {
         onChange={e => select(e)}
       />
       <div className={selectIconClass}>
-        {value === 'SHM' && <Box />}
-        {value === 'CLN' && <Clinical />}
-        {value === 'SEQ' && <Sequencing />}
-        {value === 'OTH' && <Misc />}
+        {value === 'SHM' && <SvgIcon kind="Box" width="30" height="30" />}
```
bdolly: can prob turn this into a simple map 
```js
const iconKind = {
SHM: 'Box',
CLN: 'Clinical'
...
}

<SvgIcon kind={iconKind[value]} width="30" height="30" />
```

src/components/FileDetail/FileDetail.js
```diff
@@ -52,10 +49,18 @@ const FileDetail = ({fileNode, history, match}) => {
                     <p className="FileInfo--Title">File Type:</p>
                     <div className="flex">
                       <div className="FileInfo--Icon">
-                        {fileNode.fileType === 'SHM' && <Box />}
-                        {fileNode.fileType === 'CLN' && <Clinical />}
-                        {fileNode.fileType === 'SEQ' && <Sequencing />}
-                        {fileNode.fileType === 'OTH' && <Misc />}
+                        {fileNode.fileType === 'SHM' && (
```
bdolly: can use the same map as above ☝️ maybe just move the `iconKind` into the `/common` folder as a global util

src/components/FileDetail/VersionItem.js
```diff
@@ -88,12 +63,20 @@ const VersionItem = ({
             {formatFileSize(versionNode.size, true) || 'Size Unknown'}
           </p>
           <div className="FileInfo--Icon">
-            {fileType === 'SHM' && <Box />}
-            {fileType === 'CLN' && <Clinical />}
-            {fileType === 'SEQ' && <Sequencing />}
-            {fileType === 'OTH' && <Misc />}
+            {fileType === 'SHM' && (
```
bdolly: same variable icon map idea here

src/components/Icon/Icon.js
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
```
bdolly: 👏

src/views/__tests__/FileUpload.test.js
```diff
@@ -8,6 +8,24 @@ import {mocks} from '../../../__mocks__/kf-api-study-creator/mocks';
 
 jest.mock('auth0-js');
 
+beforeAll(() => {
```
bdolly: this looks like a duplicate from __setup__, is it necessary here or does the setup logic take care of it?

XuTheBunny: this part is used to target document to render

PR #260, opened 2019-05-24
♻️ Refactor editor and upload container

XuTheBunny:
- The editor container and the upload container only contain queries and mutations.
- The upload container needs to adapt to the new requirement for uploading a new version of a file.
- The editor container needs to rewrite local cache instead of refetch query when change happens.
- The editor markup needs to adapt to the design for annotation modal on the file detail page.

### DIFF
--- a/src/components/FileAnnotation/FileEditor.js
+++ b/src/components/FileAnnotation/FileEditor.js
@@ -1,7 +1,8 @@
 import React, {useState} from 'react';
 import PropTypes from 'prop-types';
+import classes from 'classnames';
 import {withRouter} from 'react-router-dom';
-import {Button, Icon, GridContainer} from 'kf-uikit';
+import {Icon, GridContainer} from 'kf-uikit';
 import SelectElement from './SelectElement';
 import Badge from '../Badge/Badge';
 import {fileTypeDetail} from '../../common/fileUtils';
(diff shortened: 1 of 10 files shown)


src/components/FileAnnotation/FileEditor.js
```diff
@@ -18,13 +19,22 @@ const FileEditor = ({
...
+      onSubmit={e => {
+        e.preventDefault();
+        onSubmit(e);
+      }}
+      className="FileEditor"
+    >
       <GridContainer collapsed="rows">
-        {editing ? (
+        {editingName ? (
           <label className="row-1 cell-12 flex justify-between">
```
bdolly: add a `HtmlFor` prop, this will render the `for` prop that points to the input `name` for this label  so 
```
<label for="file_name_input">...</label>
<input name="file_name_input" />
```

src/components/FileAnnotation/FileEditor.js
```diff
@@ -18,13 +19,22 @@ const FileEditor = ({
...
             <input
```
bdolly: need the `name` prop as mentioned above ☝️

src/components/FileAnnotation/FileEditor.js
```diff
@@ -116,11 +119,13 @@ FileEditor.propTypes = {
...
   fileType: PropTypes.string,
   /** Action to perform when form is submitted */
-  onSubmit: PropTypes.func.isRequired,
+  onSubmit: PropTypes.func,
   /** Action to perform when name input is updated */
   onNameChange: PropTypes.func.isRequired,
   /** Function for onChange event on file type selection */
   selectFileType: PropTypes.func.isRequired,
+  /** If to hide the save/cancel buttons and badge */
+  showStatus: PropTypes.bool,
```
bdolly: @XuTheBunny I can't remember if we removed this in favor of `renderButtons`

XuTheBunny: `showStatus` is in use to conditionally render edit/save button near file name input and the badge underneath the file name input

PR #281, opened 2019-06-05
♻️ Refactor version list component

### LINKED ISSUE #266: Refactor version list component
**Updates:**
- Display version file name with click to open version detail modal
- Mark version stage as the background color of the file type icon
- Mark the latest version using a 'Latest' tag (same as in version detail modal)
- Show (part of) the version description on the second line
- Update the styles in color and font

image
image

XuTheBunny:
image

Closes #266


dankolbman:
Maybe you want to truncate description text after a certain limit?

Screen Shot 2019-06-06 at 9 26 00 AM

src/components/VersionList/VersionItem.js
```diff
@@ -0,0 +1,144 @@
...
+                    studyId,
+                    fileId,
+                    versionNode.kfId,
+                    downloadFileMutation,
+                  )
+                }
+              >
+                {versionNode.fileName}
+              </button>
+              <CopyButton
```
dankolbman: The positioning here is a little confusing as it almost implies that you are copying the file name, not the kf id.
It would be helpful if we could move this down below the avatar with the actual kf_id displayed.
Maybe something like:
Screen Shot 2019-06-06 at 10 11 59 AM

PR #289, opened 2019-06-10
♻️ Refactor upload verbiage

bdolly:
Renames all routes and references to the `file(s)` concept to use `document(s)` instead


src/containers/StudySubscriptionContainer.js
```diff
@@ -15,6 +15,7 @@ const StudySubscriptionContainer = ({
   subscribeTo,
   unsubscribeFrom,
 }) => {
+  debugger;
```
XuTheBunny: You may want to remove this line.

PR #298, opened 2019-06-12
✨ Redirect user back to the original requested page after login

### LINKED ISSUE #276: Redirect back to entry page on login
When a user enters the data tracker through some non-root url, but they are not authenticated yet, they should be sent through authentication and returned to their originally requested page.
This should be a matter of passing the original url to the LoginView which will be passed to the login provider, then back to the data tracker for re-routing.

XuTheBunny:
When a user enters the data tracker through some non-root url, but they are not authenticated yet, they should be sent through authentication and returned to their originally requested page.

Closes #276

### DIFF
--- a/src/state/auth.js
+++ b/src/state/auth.js
@@ -32,7 +32,9 @@ class Auth {
         if (authResult && authResult.accessToken && authResult.idToken) {
           localStorage.setItem('accessToken', authResult.accessToken);
           localStorage.setItem('idToken', authResult.idToken);
-          history.push('/');
+          history.push(
+            new URLSearchParams(history.location.search).get('from'),
+          );
         } else if (err) {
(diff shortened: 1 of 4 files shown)


src/state/auth.js
```diff
@@ -30,7 +32,7 @@ class Auth {
         if (authResult && authResult.accessToken && authResult.idToken) {
           localStorage.setItem('accessToken', authResult.accessToken);
           localStorage.setItem('idToken', authResult.idToken);
-          history.push('/');
+          history.push(history.location.search.slice(6));
```
bdolly: use `new URLSearchParams(history.location.search).get('from')` to grab the `from` param value, this way it's a bit more deterministic than slicing whatever string may be in the search param

PR #308, opened 2019-06-13
💄 Use Semantic for Login View

### LINKED ISSUE #300: Use Semantic for Login View
Replace the current login view with semantic

## Current
login

## Existing Implementations
We should standardize the login across apps for consistent experience. Currently, only the release coordinator utilizes a semantic ui login view:
ExistingLogin

dankolbman:
Adds Semantic and updates login view to utilize it.

Closes #300


dankolbman:
Yes, for the short term, we can simply swap over to refactoring to utilize the new components via latest stable Semantic including the vanilla styles. This will allow us to demonstrate piecewise refactoring as it would be implemented in the portal. As we advance, we can quite easily swap out both the style sheet or component library for our own forks as needed, but it seems that SteJ will likely have a large hand in this aspect, so we will simple implement the framework as is for now.
The [coordinator](https://kf-ui-release-coordinator-qa.kidsfirstdrc.org/) has already been refactored.

PR #311, opened 2019-06-13
♻️ Conditional new document form placeholder text

bdolly:
This adds unique placeholder text for each file type to help users
better understand the type of information we're looking for

### DIFF
--- a/src/forms/NewDocumentForm.js
+++ b/src/forms/NewDocumentForm.js
@@ -4,6 +4,27 @@ import {Button, GridContainer} from 'kf-uikit';
 import SelectElement from '../components/FileAnnotation/SelectElement';
 import {fileTypeDetail} from '../common/fileUtils';
 
+const plachelderText = {
+  SHM: {
+    title: 'Biospecimen shipping manifest from ...',
+    desc:
+      'Shipping manifest for biospecimens shipped from LabA to CenterB on 00/00/0000 .....',
+  },
+  CLN: {
+    title: `20XX Phenotypic data manifest for  ...`,
+    desc:
+      'Phenotypic data for 00 patients and 00 containing demographics, diagnoses, and ...',
+  },
+  SEQ: {
+    title: 'Sequencing Center manifest from [Seq. Center Name] on [date]... ',
+    desc: 'Manifest received from [Seq. Center Name] accounting for ...',
+  },
+  OTH: {
+    title: '...data sheet...',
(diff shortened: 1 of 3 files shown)


src/forms/NewDocumentForm.js
```diff
@@ -4,6 +4,27 @@ import {Button, GridContainer} from 'kf-uikit';
 import SelectElement from '../components/FileAnnotation/SelectElement';
 import {fileTypeDetail} from '../common/fileUtils';
 
+const plachelderText = {
```
dankolbman: `placeholderText`?

src/forms/NewDocumentForm.js
```diff
@@ -4,6 +4,27 @@ import {Button, GridContainer} from 'kf-uikit';
...
+    title: `20XX Phenotypic data manifest for  ...`,
+    desc:
+      'Phenotypic data for 00 patients and 00 containing demographics, diagnoses, and ...',
+  },
+  SEQ: {
+    title: 'Sequencing Center manifest from [Seq. Center Name] on [date]... ',
+    desc: 'Manifest received from [Seq. Center Name] accounting for ...',
+  },
+  OTH: {
+    title: '...data sheet...',
```
dankolbman: Is there other wording perhaps a little more helpful?  Maybe `Clinician Observation Form`

src/forms/NewDocumentForm.js
```diff
@@ -4,6 +4,27 @@ import {Button, GridContainer} from 'kf-uikit';
 import SelectElement from '../components/FileAnnotation/SelectElement';
 import {fileTypeDetail} from '../common/fileUtils';
 
+const plachelderText = {
+  SHM: {
+    title: 'Biospecimen shipping manifest from ...',
+    desc:
+      'Shipping manifest for biospecimens shipped from LabA to CenterB on 00/00/0000 .....',
```
dankolbman: Should probably be `...`

bdolly:
> The placeholder text changing on selection is a little jarring and unexpected. Is it possible to provide these same pointers in the (currently unused) item descriptions instead?

The item descriptions are meant to describe the schema for each file type whereas placeholder text is meant to encourage the user to put thought into their descriptions so that they will be useful for engineers. 

--- 

🙉 As for `jarring and unexpected`, I feel it is best to gather a greater number of opinions through testing with end-users before taking action.

PR #314, opened 2019-06-13
✨ Save user roles and groups to local storage

### LINKED ISSUE #290: add user `roles` to local client state
Throughout the application in various views we would like to conditionally show/hide elements based on a users access level/authenticated `roles`. In order to do this we need the users `accessToken` needs parsed at login and persisted into the apollo client cache

### LINKED ISSUE #312: Only show status dropdown to admin role users
The file status dropdown menu in the `Edit Document Metadata` modal show only be shown to `admin` role users as we do now want investigators to have the ability to approve their own documents. 

Non-admin users should see just the status but not be able to update it

Screen Shot 2019-06-13 at 2 00 51 PM

XuTheBunny:
Save roles and groups to local storage after authenticated and clean it up after logout

Closes #290
Closes #312


bdolly:
Closing in favor of #325

PR #315, opened 2019-06-13
💄 Use Semantic for Profile View

dankolbman:
Replace profile view with Semantic UI.

Screen Shot 2019-07-09 at 17 42 04

### DIFF
--- a/src/views/ProfileView.js
+++ b/src/views/ProfileView.js
@@ -75,7 +75,10 @@ const ProfileView = ({
                 circular
                 size="small"
                 bordered
-                src={profile.picture}
+                src={
+                  profile.picture ||
+                  'https://www.w3schools.com/css/img_avatar.png'
+                }
               />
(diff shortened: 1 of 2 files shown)


src/views/ProfileView.js
```diff
@@ -46,44 +56,69 @@ const ProfileView = ({
...
   const roles =
     decoded['https://kidsfirstdrc.org/roles'] || decoded.context.roles;
   return (
-    <GridContainer collapsed={true}>
-      <h3 className="row-1 cell-12 text-blue font-normal">Your Profile</h3>
-      <Avatar
-        className="row-2 cell-1 align-right mx-16"
-        size={100}
-        imgUrl={profile.picture}
+    <Container style={{marginTop: '30px'}}>
```
XuTheBunny: `<Container as={Segment} basic /> `

PR #316, opened 2019-06-14
💄 Version list styling updates

bdolly:
Updates the styling of the Study Document Version List to more closely match the mockups and adds emphasis to the `Latest` version list item. 

### Current 
Screen Shot 2019-06-14 at 2 04 57 PM

### Mockup
Screen Shot 2019-06-14 at 2 03 45 PM

### Updates in PR 
Screen Shot 2019-06-14 at 2 10 44 PM


src/components/VersionList/VersionItem.js
```diff
@@ -55,38 +55,33 @@ const VersionItem = ({
               </div>
             )}
             <div className="flex row-1 cell-12 lg:cell-7">
-              <button
-                className="FileVersionElement--button"
+              <p
```
dankolbman: isn't 'button' more appropriate here?

PR #318, opened 2019-06-18
💄Use Semantic for file list view

### LINKED ISSUE #302: Use Semantic for Study File List View
Replace current study file list view with Semantic.

## Current
StudyFileList

dankolbman:
This updates the study's file list view with semantic components.
Screenshot_2019-07-01 Kids First Data Tracker

Closes #302


bdolly:
@dankolbman need to move or remove the kfId tag on the right hand side as it is not part of the design spec.

PR #317, opened 2019-06-17
🚧 Semantic Refactoring

dankolbman:
This branch is a preview branch for the Semantic refactoring in #307

PR #324, opened 2019-06-19
👀 Semantic preview

dankolbman:
Preview branch to see merged state of all semantic branches.


dankolbman:
Will close now that #317 has matured.

PR #332, opened 2019-06-25
💄 Use Semantic for Study Header

### LINKED ISSUE #319: Use Semantic for Study Header
Replace the study header and navigation with semantic components.

## Current
Screen Shot 2019-06-18 at 4 07 05 PM

dankolbman:
Replaces the study nav bar and study header with Semantic components.
Also reorganizes the NavBar component to be the StudyNavBar as it is only used in the study view.

Screen Shot 2019-06-25 at 11 56 41 AM

Closes #319


dankolbman:
@bdolly the header is nested instead a standard `<Container />` which will maintain a consistent max width. It will be aligned with the file list once the file list has been refactored.

As for the header spacing, alignment, and size, this usage is correct as far as headers and subheaders go in semantic.

bdolly:
> @bdolly the header is nested instead a standard `<Container />` which will maintain a consistent max width. It will be aligned with the file list once the file list has been refactored.
> 
> As for the header spacing, alignment, and size, this usage is correct as far as headers and subheaders go in semantic.

@dankolbman right but as the designer of the application I'm saying that we need to adjust those default values as the original ones were designed and implemented for a purpose

PR #350, opened 2019-07-10
💄 Use Semantic for File Annotation view

### LINKED ISSUE #304: Use Semantic for New File View
Replace current new file form with Semantic.

## Current
NewFile

XuTheBunny:
image
image

Semantic components used:
- Icon
- Header
- Button
- Label
- Segment
- Form
- Dropdown
- Radio

Closes #304


src/components/FileAnnotation/SelectElement.js
```diff
@@ -1,42 +1,45 @@
...
-        inverted
-        color={selected ? 'blue' : null}
-      />
-      <div>
-        <p className="m-0 font-bold font-sm">{fileTypeDetail[value].title}</p>
-        <span className="font-normal text-grey text-xs">
-          {fileTypeDetail[value].description}
-        </span>
-      </div>
+    <label>
```
bdolly: 🔥 label elements should be text only with no child elements http://html5doctor.com/element-index/#label

src/components/FileAnnotation/SelectElement.js
```diff
@@ -1,42 +1,45 @@
...
+      <Segment
```
bdolly: let's make this the top-level element and add an `onClick` prop that calls `select` this way the entire element will act as a radio button with the added bonus of being easy to use on mobile

src/components/FileAnnotation/SelectElement.js
```diff
@@ -1,42 +1,45 @@
...
+        compact
+        clearing
+        color={selected && 'blue'}
+        className="selection--card"
+      >
+        <Header as="h5" floated="left">
```
bdolly: 🔥 remove all the `Header` elements from the component as headings should typically be text only

src/components/FileAnnotation/SelectElement.js
```diff
@@ -1,42 +1,45 @@
...
+          <Radio
+            name={value}
```
bdolly: make the name something like `radioGroup` or `fileTypeGroup` https://react.semantic-ui.com/addons/radio/#types-radio-group

src/components/FileAnnotation/SelectElement.js
```diff
@@ -1,42 +1,45 @@
...
+            value={value}
+            checked={selected}
+            onChange={e => select(e)}
```
bdolly: add `label` as a render prop and pass the icon to it, this will replace the default browser checkbox with the icon 
```
label={()=> ( icon goes here))
```

src/components/FileAnnotation/SelectElement.js
```diff
@@ -1,42 +1,45 @@
...
+          />
+        </Header>
+        <Header as="h5" floated="left">
+          <Icon
+            name={`${fileTypeDetail[value].icon || 'question'}`}
```
bdolly: no need to make this conditional as this `icon` should always be set

src/components/FileAnnotation/SelectElement.js
```diff
@@ -1,42 +1,45 @@
...
+            size="large"
+            bordered
+            circular
+            inverted
+            color={selected ? 'blue' : null}
+          />
+          <Header.Content>
```
bdolly: ♻️ let's replace all this semantic `Header.Content` stuff with just a regular inline block div with some html headings in it and adjust the margins and padding to match the designs
```
div.inline-block 
  h5 title
  p > small  description
```

src/App.css
```diff
@@ -101,3 +101,9 @@ body {
   padding-top: 7px !important;
   padding-bottom: 7px !important;
 }
+
+.selection--card {
+  cursor: pointer;
+  width: 100%;
+  margin-bottom: 5px !important;
```
bdolly: nest this class in here to help better align the icon to the text 
```css
.ui.checkbox{
  vertical-align: bottom;
}
```

src/forms/EditDocumentForm.js
```diff
@@ -18,72 +17,53 @@ const EditDocumentForm = ({
...
+          value={versionStatus}
+          placeholder="Choose an option"
+          onChange={(e, {value}) => {
+            onVersionStatusChange(value);
+          }}
+        />
+      </Form.Field>
+      <Form.Field required>
+        <label htmlFor="file_type">Document Type:</label>
+        <fieldset name="file_type">
```
bdolly: 🔥 `fieldset` no need for the nesting here

src/forms/EditDocumentForm.js
```diff
@@ -18,72 +17,53 @@ const EditDocumentForm = ({
...
           {Object.keys(fileTypeDetail).map(item => (
             <SelectElement
```
bdolly: make each SelectElement in you fileTypeDetail iteration a Form.Field

Let's also update the `select` prop to just use the value and not the event since we called it with the raw value in the SelectElement component
```
  < Form.Field>
      <SelectElement select={val => onFileTypeChange(val}  />
   < /Form.Field>
```

src/components/FileAnnotation/SelectElement.js
```diff
@@ -1,43 +1,44 @@
...
-      />
-      <Icon
-        name={`${fileTypeDetail[value].icon || 'question'}`}
-        size="large"
-        bordered
-        circular
-        inverted
-        color={selected ? 'blue' : null}
+        onChange={select}
+        label={() => (
```
dankolbman: You want to make sure there is text in the label for screen readers. I think you could also reduce the amount of markup in the label by using some of the semantic classes which also means you can get rid of some of the css defs:
```
label={() => (
  <Header as="span" size="small">
    <Icon
      name={fileTypeDetail[value].icon}
      size="large"
      bordered
      circular
      inverted
      color={selected ? 'blue' : 'black'}
    />
    <Header.Content as="span">
      {fileTypeDetail[value].title}
      <Header.Subheader as="small">
        {fileTypeDetail[value].description}
      </Header.Subheader>
    </Header.Content>
  </Header>
)}
```

src/forms/EditDocumentForm.js
```diff
@@ -17,98 +16,107 @@ const EditDocumentForm = ({
...
-      <GridContainer>
-        <label
-          className="Form--Label row-1 cell-3 text-right Form--Label-required"
-          htmlFor="file_name"
+  const [newFileType, setFileType] = useState();
+  const [newFileName, setFileName] = useState();
+  const [newFileDescription, setFileDescription] = useState();
+  const [onUploading, setUploading] = useState(false);
+  const options = [];
+  Object.keys(versionState).forEach(function(e) {
```
dankolbman: We typically opt for immutable variables when possible. Even though the `options` are declared const, it's still mutable because it's an `Array` and then immediately mutated in the `forEach`.
Instead, consider using a `map`, as is more typical in our code base:
```
const options = Object.keys(versionState).map(state => ({key: state, ...}))
```

src/forms/EditDocumentForm.js
```diff
@@ -17,98 +16,107 @@ const EditDocumentForm = ({
...
+    options.push({
+      key: e,
+      value: e,
+      text: versionState[e].title,
+      content: (
```
dankolbman: I _think_ that `content: <Badge state={state} />` is valid here without the extra `<button />`

src/components/FileAnnotation/SelectElement.js
bdolly: I think the above solution ☝️ is too much unnecessary nesting just for the sake of using the Semantic Header component. 
  
This would render the below markup, which is spancity, spanville USA and hard to style because then the size of the `description`  would need to be `span > span >small {   }` that could simply be inherited from a more semantically meaningul tag like a `<p>` 
```html
<span>
    <i />
   <span>
     {title}
        <small>{description}</small>
  </span>
</span>
```

dankolbman: @bdolly the tags can be changed around however we want. This is also not just for the sake of using the Header component, but rather for the sake of re-using an existing design pattern for Icon-header-subheaders that has already been established. 
Regardless, it still seems preferable to the previous version:
```
<div >
    <i />
</div>
<div>
   <p>
        <b>{title}</b>
    </p>
   <p>
        <small>{description}</small>
    </p>
</div>
```

bdolly: preferable why?

dankolbman: The previous version is less preferable because the `<input />` does not have its label as an immediate sibling element.

bdolly: ahh gotcha 👍

src/forms/EditDocumentForm.js
```diff
@@ -17,98 +16,95 @@ const EditDocumentForm = ({
...
+      onSubmit={ev => {
+        if (handleSubmit) {
+          setUploading(true);
+          ev.preventDefault();
+          handleSubmit(newFileName, newFileType, newFileDescription);
+        }
+      }}
+    >
+      <Form.Field required>
+        <label>Document Title:</label>
```
bdolly: needs an `htmlFor` prop

src/forms/EditDocumentForm.js
```diff
@@ -17,98 +16,95 @@ const EditDocumentForm = ({
...
-        <label
-          className="Form--Label row-3 cell-3 text-right Form--Label-required"
-          htmlFor="file_type"
-        >
-          Document Type:
-        </label>
-        <fieldset name="file_type" className="row-3 cell-9">
-          {Object.keys(fileTypeDetail).map(item => (
+      </Form.Field>
+      {versionStatus && (
```
bdolly: is this scoped to show only `Admin` role users?

src/forms/EditDocumentForm.js
```diff
@@ -17,98 +16,95 @@ const EditDocumentForm = ({
...
+          onChange={e => {
+            onDescriptionChange
+              ? onDescriptionChange(e)
+              : setFileDescription(e.target.value);
+          }}
         />
-      </GridContainer>
-    </form>
+      </Form.Field>
+      {submitButtons &&
```
bdolly: be sure to add `submitButtons` to your propTypes

src/views/NewDocumentView.js
```diff
@@ -41,31 +40,55 @@ const NewDocumentView = ({match, history, location, createDocument}) => {
...
-      </div>
-      <section className="study-file-list cell-12 row-4">
-        <NewDocumentForm
-          handleSubmit={handleSubmit}
-          handleCancel={() =>
-            history.push(`/study/${match.params.kfId}/documents`)
-          }
-          errors={errors}
+    <Container as={Segment} vertical basic>
+      {errors ? (
```
bdolly: the error message should not be part of a ternary with the message. It should just be conditionally rendered under the `Tell us about...` content

src/views/NewDocumentView.js
```diff
@@ -41,31 +40,55 @@ const NewDocumentView = ({match, history, location, createDocument}) => {
...
+        <Message
+          negative
+          icon="warning circle"
+          header="Error"
+          content={errors}
+        />
+      ) : (
+        <Message
```
bdolly: move this content to a `h3` header and `p` description instead of a `message` as it's essentially the page sub-heading with a description below it

src/views/NewDocumentView.js
```diff
@@ -41,31 +40,55 @@ const NewDocumentView = ({match, history, location, createDocument}) => {
...
+          icon="edit"
+          header="Tell us about your study document"
+          content="Help ensure the fastest processing and harmonization of your study by telling us about the contents of your uploaded document. This helps our engineers accurately interpret your data."
         />
-      </section>
-    </GridContainer>
+      )}
+      <Segment.Group>
+        <Segment>
+          <b>Uploaded File: {location.state.file.name}</b>
```
bdolly: let's make this an `h4` with the `Upload File:` being bolded

src/forms/EditDocumentForm.js
XuTheBunny: Will address in separate pr.

PR #352, opened 2019-07-11
💄 Use Semantic for error message

### LINKED ISSUE #310: 💄create better error page design and handling
Screen Shot 2019-06-13 at 10 43 45 AM
 Right now we just cough up the raw graphql error which is only helpful for developers. A more stylized and thoughtful design needs to be made for an error page prompting users to contact support if the error continues

XuTheBunny:
- Study list view
image
image

- Token view
image

- File list view
image

- Profile view
image

- NavBar view
image

- VersionUpload modal
image

- File detail view
image

- File upload target
image

Closes #310

### DIFF
--- a/src/components/FileUpload/FileUploadTarget.js
+++ b/src/components/FileUpload/FileUploadTarget.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import {Button, Header, Icon, Segment} from 'semantic-ui-react';
+import {Button, Header, Icon, Segment, Message} from 'semantic-ui-react';
 
 const FileUploadTarget = props => {
   const {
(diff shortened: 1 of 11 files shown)


src/components/FileUpload/FileUploadTarget.js
```diff
@@ -42,9 +42,12 @@ const FileUploadTarget = props => {
...
         />
         {error && error.graphQLErrors && (
-          <Header as="h4">
-            Failed to upload: {error.graphQLErrors.map(err => err.message)}
-          </Header>
+          <Message
+            negative
+            icon="warning circle"
+            header="Failed to upload:"
+            content={error.graphQLErrors.map(err => err.message)}
```
dankolbman: Maybe `error.graphQLErrors.join(', ')` would be better so you have a formatted string instead of an array of strings.

PR #353, opened 2019-07-11
💄Feature/semantic login styling

bdolly:
Makes a number of design improvements to the Login View 

Screen Shot 2019-07-11 at 4 49 38 PM


src/components/LoginCard/LoginCard.js
```diff
@@ -1,11 +1,12 @@
 import React from 'react';
 import GoogleLogin from 'react-google-login';
-import {GOOGLE_APP_ID, EGO_API} from '../../common/globals';
```
dankolbman: It looks like maybe prettier didn't pick up the prettier config file in the root of the project, or maybe we are missing some configuration setting?

src/components/LoginCard/LoginCard.js
```diff
@@ -40,37 +41,37 @@ const onFailure = repsonse => {
...
           onClick={() => auth.login()}
-          positive
-          icon
-          labelPosition="right"
+          size="large"
+          color="black"
         >
-          Login with Auth0
-          <Icon name="chevron right" />
+          <img className="auth0-logo" src={Auth0_logo} alt="Auth0 logo" />
```
dankolbman: It looks like we can get the shield logo alone [here](https://styleguide.auth0.com/resources/logos). We could use this sans the auth0 text to have things align more naturally and be more consistent with all the other [buttons](https://styleguide.auth0.com/resources/logos).
This would be a good extension to make to the Semantic `<Button />` component.

src/App.css
```diff
@@ -16,15 +16,28 @@ body {
...
+    .center.column{
+      min-width: 300px !important;
+    }
 }
 
-.Card--Login {
-  @extend .Card;
-  width: 620px;
+.ui.button.button--login{
+    margin-bottom: 10px;
```
dankolbman: We can use Semantic's spacing to accomplish the same thing by setting `<Segment padded />`, this way we can reproduce the login component between apps without copy+pasting css and it will insure we have consistent levels of spacing when using Segments.

src/App.css
```diff
@@ -16,15 +16,28 @@ body {
...
+    .auth0-logo{
+      width: 70px;
+    }
+
+  }
+  .ui.buttons .or.button--login{
```
dankolbman: It looks like this is unused?

src/App.css
```diff
@@ -16,15 +16,28 @@ body {
...
+      min-width: 300px !important;
+    }
 }
 
-.Card--Login {
-  @extend .Card;
-  width: 620px;
+.ui.button.button--login{
+    margin-bottom: 10px;
+    .auth0-logo{
```
dankolbman: When the auth0 image changes to the shield icon, this shouldn't be needed.

src/components/LoginCard/LoginCard.js
dankolbman: The `shield` icon that ships with semantic is also decently close and has consistent styling with other icons.

src/views/LoginView.js
```diff
@@ -1,34 +1,45 @@
...
+            <Segment.Group raised >
+              <Segment>
+                <Header as="h3"> Log In</Header>
+                <LoginCard />
+              </Segment>
+              <Segment secondary>
+                <p>
+                  New to Kids First Data Tracker?
+                  &nbsp;
+                  <Responsive as="br" maxWidth={1770} />
```
dankolbman: Could use the predefined breakpoints here: `Responsive.onlyLargeScreen.minWidth`

src/components/LoginCard/LoginCard.js
bdolly: decently close is not good enough for me

src/App.css
bdolly: the LoginCard (probably need to more accurately name the component) uses it

src/App.css
bdolly: 👎 not using the shield icon in the design

dankolbman:
Can you update pr title, assign yourself, assign labels, fix testes, and rebase on master?

PR #363, opened 2019-07-15
💄 Semantic refactor for avatar image component

### LINKED ISSUE #360: User avatar doesn't render when null
The user avatar in the label on the File Detail View doesn't render if the `creator` returned by the api is `null`

Screen Shot 2019-07-12 at 10 38 53 AM

XuTheBunny:
1. With creator picture
Screen Shot 2019-07-15 at 3 08 35 PM

2. With creator but no picture
Screen Shot 2019-07-15 at 3 08 15 PM

3. Without creator
Screen Shot 2019-07-15 at 3 09 01 PM

4. Without creator and no createdAt time
Screen Shot 2019-07-15 at 3 09 13 PM

5. Tokens with creator picture
Screen Shot 2019-07-15 at 3 06 54 PM

6. Tokens without creator picture
Screen Shot 2019-07-15 at 3 07 10 PM

Close #360


src/components/AvatarTimeAgo/AvatarTimeAgo.js
```diff
@@ -6,18 +6,16 @@ import TimeAgo from 'react-timeago';
  * Displays time ago and avatar in one label
  */
 const AvatarTimeAgo = ({showUsername, creator, createdAt, size}) => {
-  if (showUsername) {
+  const picUrl =
+    creator && creator.picture
+      ? creator.picture
+      : 'https://www.w3schools.com/css/img_avatar.png';
```
dankolbman: Since we use this in a couple locations, maybe it's better to pull it out a configuration variable maybe somewhere under `src/common/`?

dankolbman: It might even be better to bundle our own under `src/assets/`.

PR #389, opened 2019-07-30
✨ Filter files by approval status/type, sort by created/modified date, and search by file title/description

### LINKED ISSUE #103: Make files list sortable
To assist with large numbers of files, ordering of files will be helpful.

Order by:
- created at
- file name

### LINKED ISSUE #180: Make files list filterable
To assist with large numbers of files, filtering of files will be helpful.

Filter by:
- file type
- file state

### LINKED ISSUE #181: Search files list
The files list should be easier to find files by allowing the user to search the files in the study by name or description. This may be done through a simple text box above the list that will live-filter the list based on whether the filename or description contains the entered text.

XuTheBunny:
Default:
image

Filter by approval status:
image

Filter by file type:
image

Sort by date:
image

Seach:
image

Combo:
image

Closes #103 
Closes #180
Closes #181

### DIFF
--- a/src/common/fileUtils.js
+++ b/src/common/fileUtils.js
@@ -130,3 +130,14 @@ export const createDateSort = (a, b) => {
     ? -1
     : 0;
 };
+
+// Sort files by modified date
+export const modifiedDateSort = (a, b) => {
+  return Date.parse(fileLatestDate(fileSortedVersions(a.node))) >
+    Date.parse(fileLatestDate(fileSortedVersions(b.node)))
+    ? 1
+    : Date.parse(fileLatestDate(fileSortedVersions(b.node))) >
+      Date.parse(fileLatestDate(fileSortedVersions(a.node)))
+    ? -1
+    : 0;
+};
--- a/src/common/fileUtils.js
+++ b/src/common/fileUtils.js
@@ -119,3 +119,14 @@ export const fileLatestStatus = fileNode => {
 export const lengthLimit = (text, limit) => {
   return text.length > limit ? text.substring(0, limit) + '...' : text;
 };
+
+// Sort files by created date
+export const createDateSort = (a, b) => {
+  return Date.parse(fileOldestDate(fileSortedVersions(a.node))) >
+    Date.parse(fileOldestDate(fileSortedVersions(b.node)))
+    ? 1
+    : Date.parse(fileOldestDate(fileSortedVersions(b.node))) >
+      Date.parse(fileOldestDate(fileSortedVersions(a.node)))
+    ? -1
+    : 0;
+};
--- a/src/common/fileUtils.js
+++ b/src/common/fileUtils.js
@@ -80,10 +80,19 @@ export const fileSortedVersions = fileNode => {
   } else return [];
 };
 
-// Get the lastest date from the latest version of the file
+// Get the file modified date from the latest version of the file
 export const fileLatestDate = sortedVersions => {
   if (sortedVersions.length > 0) {
     return sortedVersions[0].node.createdAt;
+  } else {
+    return null;
+  }
+};
+
+// Get the file created date from the oldest version of the file
+export const fileOldestDate = sortedVersions => {
+  if (sortedVersions.length > 0) {
+    return sortedVersions.reverse()[0].node.createdAt;
   } else {
     return null;
   }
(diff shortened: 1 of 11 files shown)


src/components/FileList/FileList.js
```diff
@@ -20,17 +31,164 @@ const FileList = ({fileList, studyId}) => {
...
+    if (
+      sortMethod === '' &&
+      typeFilterStatus === '' &&
+      approvalFilterStatus === ''
+    ) {
+      sortedList = fileList
+        .filter(obj => fileLatestStatus(obj.node) === 'CHN')
+        .concat(fileList.filter(obj => fileLatestStatus(obj.node) !== 'CHN'));
+    } else {
+      if (sortMethod === 'createDate') {
```
dankolbman: It may be cleaner if you move these sort methods to utility functions and map the method to the sort function:
```
const sortsMethods = { 'createDate': sortOldestVersion, 'modifiedDate': sortLatestDate};
const sortedList = fileList.sort(sortMethods[sortMethod])
```

src/components/FileList/FileList.js
```diff
@@ -20,17 +31,164 @@ const FileList = ({fileList, studyId}) => {
...
+        />
+      </Segment>
+      <Segment
+        className="noMargin noHorizontalPadding"
+        basic
+        compact
+        floated="right"
+      >
+        <span className="smallLabel">Sorted by:</span>
+        <Dropdown
```
dankolbman: Clicking the dropdowns will automatically select the first element when clicking out of the select box. This is a bit obnoxious as I don't usually expect to have a filter applied if I 'cancel' the select. I think this can be disabled through setting `selectOnBlur` to `false`.

src/common/fileUtils.js
```diff
@@ -110,3 +119,39 @@ export const fileLatestStatus = fileNode => {
 export const lengthLimit = (text, limit) => {
   return text.length > limit ? text.substring(0, limit) + '...' : text;
 };
+
+// Sort files by created date
+export const createDateSort = (fileList, sortDirection) => {
```
dankolbman: Instead of taking and returning a list of values, maybe just return the function?
Something where we can do `[file1, file2, ...].sort(createDateSort)`

src/components/FileList/FileList.js
```diff
@@ -20,17 +31,189 @@ const FileList = ({fileList, studyId}) => {
...
+  ];
+  const searchOptions = [
+    {key: 'name', text: 'Title', value: 'name'},
+    {key: 'description', text: 'Description', value: 'description'},
+  ];
+
+  const sortedFileList = () => {
+    var sortedList = [];
+    sortedList = fileList;
+    if (
```
dankolbman: If you write the sort functions so that they could be passed to `.sort()` and store them in an object, I think you could get rid of all of these if/elses which are a little tricky to read through:
```
const sortFuncs = {'createDate': createDateSort, 'default': defaultSort}
const sortedList = fileList.sort(sortFuncs[sortMethod] || sortFuncs.default)
```
```

File: src/components/FileList/FileList.js [hunk 5]
```diff
@@ -20,17 +31,189 @@ const FileList = ({fileList, studyId}) => {
...
+          obj => obj.node.fileType === typeFilterStatus,
+        );
+      }
+      if (approvalFilterStatus !== '') {
+        sortedList = sortedList.filter(
+          obj => fileLatestStatus(obj.node) === approvalFilterStatus,
+        );
+      }
+    }
+    if (searchString !== '') {
```
dankolbman: I have a feeling there is a way to avoid checking the value of the filters. If you filter objects by an empty string, does it return all results, or none?

src/components/FileList/FileList.js
```diff
@@ -20,17 +31,189 @@ const FileList = ({fileList, studyId}) => {
...
+          <Icon name={'sort content ' + sortDirection} />
+        </Button>
+      </Segment>
+      <Segment
+        className="noMargin noHorizontalPadding"
+        basic
+        compact
+        floated="right"
+      >
+        <Input
```
dankolbman: Out of personal preference, and given that we don't have a huge number of files and descriptions, I think we could instead filter on both fileName and description by concatinating both of them and simply returning any that have the search term in them.
That way, we don't need to be bothered with selecting which field we want to search on.

PR #390, opened 2019-07-31
✨ Search study by its name / short name from study list screen

### LINKED ISSUE #152: Study text filter
So for #151 I ended up just doing ctrl-f a lot to find the studies of interest. Could we have a filter box to type in that searches against the study entity, especially as we continue to get more studies in the future and might consider paginating?

### LINKED ISSUE #153: Make studies easier to sort through
There is a need to make sorting through our vast amount of studies from the administrator's standpoint more easy. The investigator's should be fine with only 1-3 cards available to them, but admins will have all studies available to them, cluttering the view and making it difficult to pick out the study they need.

Current ideas for improvement are:
- text filtering #152 
- 'favorite' studies, allow a user to pick studies that will be displayed for them at the top of the page
- alternate table view - easier sorting and filtering

XuTheBunny:
Screen Shot 2019-07-31 at 5 08 43 PM
Screen Shot 2019-07-31 at 5 08 29 PM

Closes #152
Closes #153

### DIFF
--- a/src/components/StudyList/StudyList.js
+++ b/src/components/StudyList/StudyList.js
@@ -23,6 +23,7 @@ const HeaderSkeleton = () => (
  */
 const StudyList = ({studyList, loading, activeView = 'grid'}) => {
   const [view, setView] = useState(activeView);
+  const [searchString, setSearchString] = useState('');
 
   if (loading) {
     return (
--- a/src/components/StudyList/StudyList.js
+++ b/src/components/StudyList/StudyList.js
@@ -34,6 +34,19 @@ const StudyList = ({studyList, loading, activeView = 'grid'}) => {
     );
   }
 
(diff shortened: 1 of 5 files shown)


src/components/StudyList/StudyList.js
```diff
@@ -33,19 +41,40 @@ const StudyList = ({studyList, loading, activeView = 'grid'}) => {
...
+          iconPosition="left"
+          icon="search"
+          placeholder="Search by Study Name"
+          onChange={(e, {value}) => {
+            setSearchString(value);
+          }}
+          value={searchString}
+        />
         <ToggleButtons
+          floated="left"
```
dankolbman: Does this need to be floated?

XuTheBunny: Removed

PR #403, opened 2019-08-19
✨ Add form page for creating new study

XuTheBunny:
image

### DIFF
--- a/src/Routes/index.js
+++ b/src/Routes/index.js
@@ -26,7 +26,7 @@ const Routes = () => (
     </Switch>
     <PrivateRoute path="/profile" component={ProfileView} />
     <PrivateRoute path="/study/:kfId(SD_\w{8})/" component={NavBarView} />
-    <PrivateRoute exact path="/study/new-study" component={NewStudyView} />
+    <AdminRoute exact path="/study/new-study" component={NewStudyView} />
     <Switch>
       <PrivateRoute exact path="/" component={StudyListView} />
       <PrivateRoute
--- a/src/Routes/index.js
+++ b/src/Routes/index.js
@@ -15,6 +15,7 @@ import {
   CavaticaProjectsView,
   FileDetailView,
   ProfileView,
+  NewStudyView,
 } from '../views';
 const Routes = () => (
   <Fragment>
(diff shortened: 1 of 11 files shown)


src/components/StudyList/StudyList.js
```diff
@@ -60,6 +62,15 @@ const StudyList = ({studyList, loading, activeView = 'grid'}) => {
         <Header as="h1">Your Studies</Header>
       </Grid.Column>
       <Grid.Column width={8} textAlign="right">
+        <Button
```
dankolbman: Do you want to hide this for non-admin users?

src/components/StudyList/StudyList.js
```diff
@@ -114,4 +132,4 @@ StudyList.defaultProps = {
   activeView: 'grid',
 };
 
-export default StudyList;
+export default compose(graphql(MY_PROFILE, {name: 'user'}))(StudyList);
```
dankolbman: You shouldn't need the `compose` here since there's only one `graphql` query.

src/Routes/index.js
```diff
@@ -23,7 +24,8 @@ const Routes = () => (
       <Route path="/" render={() => <Header />} />
     </Switch>
     <PrivateRoute path="/profile" component={ProfileView} />
-    <PrivateRoute path="/study/:kfId/" component={NavBarView} />
+    <PrivateRoute path="/study/:kfId(SD_\w{8})/" component={NavBarView} />
+    <PrivateRoute exact path="/study/new-study" component={NewStudyView} />
```
dankolbman: Maybe this should be an admin route? https://github.com/kids-first/kf-ui-data-tracker/blob/ace8fc90d2926fefdf91a27801be896895e2a9b1/src/Routes/AdminRoute.js#L27-L34

src/views/NewStudyView.js
```diff
@@ -0,0 +1,97 @@
...
+      })
+      .catch(err => {
+        setNewStudyError(err.message);
+        setStudyCreating(false);
+      });
+  };
+
+  return (
+    <Container as={Segment} vertical basic>
+      {isAdmin ? (
```
dankolbman: If you use an AdminRoute for this view, you won't have to have this big conditional render in here.

PR #435, opened 2019-09-03
♻️ Refactor project list showing unlink button for admin user

XuTheBunny:
image


src/components/CavaticaProjectList/CavaticaProjectItem.js
```diff
@@ -0,0 +1,110 @@
...
+    <List.Item>
+      <Icon
+        name={
+          projectNode.projectType === 'DEL'
+            ? 'paper plane outline'
+            : 'sliders horizontal'
+        }
+      />
+      <List.Content>
+        {unlinkProject && (projectNode.study || studyId) ? (
```
dankolbman: Could we show the study that it's linked to? That would be helpful to know when we're viewing the list outside a single study's context.

PR #447, opened 2019-09-05
✨ Support markdown for study description

XuTheBunny:
image
image
image
image

### DIFF
--- a/src/components/StudyInfo/StudyInfo.js
+++ b/src/components/StudyInfo/StudyInfo.js
@@ -1,10 +1,19 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import TimeAgo from 'react-timeago';
-import {List, Grid, Header, Segment, Button} from 'semantic-ui-react';
+import {
+  List,
+  Grid,
+  Header,
+  Segment,
+  Button,
+  Image,
+  Table,
+} from 'semantic-ui-react';
 import EventList from '../EventList/EventList';
 import FileSimpleList from './FileSimpleList';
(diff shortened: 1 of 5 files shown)


src/components/StudyInfo/StudyInfo.js
```diff
@@ -127,22 +137,43 @@ const StudyInfo = ({studyNode, setShowModal, unlinkProject}) => {
...
         <Grid.Column
           mobile={16}
           tablet={16}
           computer={longDescription ? 16 : 8}
         >
-          <Segment padded>
+          <Segment
+            padded
+            attached={
+              studyNode.description && studyNode.description.length > 3500
```
dankolbman: Now that I think of it, it may be better to not try to do the read more/less based on length as we allow images, tables, and other nodes that may have inconsistent line height.
Instead, can we just always show the read more/less regardless of the content length which simply switches off the max height class?

src/components/StudyInfo/StudyInfo.js
```diff
@@ -127,22 +137,43 @@ const StudyInfo = ({studyNode, setShowModal, unlinkProject}) => {
...
         >
-          <Segment padded>
-            <Header as="h3">Description</Header>
-            {studyNode.description && studyNode.description.length > 0 ? (
-              <p>{studyNode.description}</p>
-            ) : (
-              <Header disabled textAlign="center" as="h4">
-                No study description available.
-              </Header>
+          <div>
```
dankolbman: Can you use a fragment here?

PR #454, opened 2019-09-09
♻️ Refactor create project and list project rendering

XuTheBunny:
**Improvements:**
- Ensure error message are handled and passing valid props (on success case and error case)
- Ensure the page data (cache) are updated after mutation
- Emphasize less on the deleted projects 
- Hide redundant info for clear appearance 

**UI changes:**
Cavatica Project card on Study Basic Info page:
Before:
image
After:
image

Cavatica Projects page:
Before:
image
After:
image

### DIFF
--- a/src/components/CavaticaProjectList/CavaticaProjectItem.js
+++ b/src/components/CavaticaProjectList/CavaticaProjectItem.js
@@ -1,7 +1,49 @@
 import React from 'react';
 import {Link} from 'react-router-dom';
-import {List, Icon, Button, Popup, Divider} from 'semantic-ui-react';
+import {List, Icon, Button, Popup, Divider, Header} from 'semantic-ui-react';
 import TimeAgo from 'react-timeago';
+
+const ProjectAttributes = ({projectNode}) => (
+  <List bulleted horizontal>
+    <List.Item>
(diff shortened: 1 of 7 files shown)


bdolly:
@XuTheBunny 
Screen Shot 2019-09-10 at 1 37 01 PM
The button pattern for `New Project` and `Link Project` is confusing since it looks more like a toggle than separate buttons. It could easily be confused with the pattern below
Screen Shot 2019-09-10 at 1 28 54 PM
 
Could we make them separate buttons where the `New Project` looks like the **Add Study** ☝️ and `Link Project` is solid with maybe the _linkify_ icon

src/components/CavaticaProjectList/CavaticaProjectItem.js
```diff
@@ -1,36 +1,96 @@
...
+      <code>{projectNode.projectId}</code>
+    </List.Item>
+  </List>
+);
+
+const ProjectLink = ({projectNode, disableLink}) => (
+  <List.Header
+    as="a"
+    target="_blank"
+    href={
```
dankolbman: can you do `!disableLink && 'http://...'` here?

PR #459, opened 2019-09-10
💄Study files filter bar style update

bdolly:
Updates the file list filters to be it's own component and styled as a semantic menu in order to provide a more clear visual separation between list actions and list items in the StudyFilesListView
## Updated
Screen Shot 2019-09-10 at 1 08 40 PM

## Previous
Screen Shot 2019-09-10 at 1 11 55 PM

### DIFF
--- a/src/components/ListFilterBar/ListFilterBar.js
+++ b/src/components/ListFilterBar/ListFilterBar.js
@@ -1,0 +1,190 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import {
+  Icon,
+  Segment,
+  Button,
+  Dropdown,
+  Input,
+  Divider,
+} from 'semantic-ui-react';
+import Badge from '../Badge/Badge';
+import {
+  fileLatestStatus,
+  versionState,
+  fileTypeDetail,
+  createDateSort,
+  modifiedDateSort,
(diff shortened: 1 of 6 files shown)


src/components/ListFilterBar/ListFilterBar.js
```diff
@@ -0,0 +1,164 @@
...
+        {' ' + fileTypeDetail[type].title}
+      </small>
+    ),
+  }));
+
+  const sortOptions = [
+    {
+      key: 'createDate',
+      value: 'createDate',
+      text: 'Create date',
```
dankolbman: 'Created Date`? `Creation Date`?

src/components/ListFilterBar/ListFilterBar.js
```diff
@@ -0,0 +1,164 @@
...
+      obj =>
+        obj.node.name.toLowerCase().includes(searchString.toLowerCase()) ||
+        obj.node.description.toLowerCase().includes(searchString.toLowerCase()),
+    );
+    return sortedList;
+  };
+
+  return (
+    <>
+      <Menu size="small">
```
dankolbman: Did you consider using a form group with inline fields here? It may help with some gutters and making it more consistent with other input fields.

PR #466, opened 2019-09-16
♻️ Refactor project button and link rendering

XuTheBunny:
**Improvements:**
- Filter out deleted projects from link project options dropdown.
- Show unlink project button when the project is deleted but still linked
- Show project name as the regular header when link is disbaled and hide link icon
- Add proptypes checking for CavaticaProjectItem component
- Minor fix on prop type error

**UI changes:**
Cavatica Project card on Study Basic Info page:
Before:
image
After:
image

Cavatica tab under the study:
Before:
image
After:
image

Link project modal:
Before:
image
After:
image

Cavatica project list page for ADMIN:
Before:
image
After:
image

### DIFF
--- a/src/components/CavaticaProjectList/CavaticaProjectItem.js
+++ b/src/components/CavaticaProjectList/CavaticaProjectItem.js
@@ -46,6 +46,49 @@ const StudyLink = ({study, hideLink}) => {
   }
 };
 
+const UnlinkButton = ({unlinkProject, study, projectId}) => (
+  <Popup
+    trigger={
+      <Button
+        basic
+        negative
+        floated="right"
+        size="mini"
+        icon="unlink"
+        content="UNLINK"
+        className="ml-10"
+        onClick={e => e.stopPropagation()}
+      />
+    }
+    header="Are you sure?"
+    content={
+      <>
+        This will unlink the project from its study. It may always be linked
+        back later.
+        <Divider />
+        <Button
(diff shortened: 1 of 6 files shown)


src/components/CavaticaProjectList/CavaticaProjectItem.js
```diff
@@ -45,17 +46,71 @@ const StudyLink = ({study, hideLink}) => {
   }
 };
 
+const UnlinkButton = ({unlinkProject, study, projectId}) => {
+  if (unlinkProject && study) {
```
dankolbman: It's probably a better idea to move the conditional rendering here into the parent. There's no need to include the component if it's not going to be displayed.

dankolbman: You may want to add proptypes for this component if you're concerned about what values you're getting for it.

src/components/CavaticaProjectList/CavaticaProjectItem.js
```diff
@@ -45,17 +46,71 @@ const StudyLink = ({study, hideLink}) => {
...
+    return null;
+  }
+};
+
 const CavaticaProjectItem = ({
   projectNode,
   unlinkProject,
-  studyId,
   disableLink,
+  hideStudy,
```
dankolbman: It would be nice to have proptypes with descriptions for each of these props.

PR #468, opened 2019-09-18
✨ Add event logs tab view for ADMIN user with type filter and auto refresh

XuTheBunny:
**Feature:**
- For ADMIN user, they can view a list of event logs for the study
- For ADMIN user, they can filter event logs by event type
- Event logs refreshed every 30 seconds when on logs page

**UI changes:**
Add Logs tab under the study:
Display logs:
image
Applied filter:
image
Show empty page for not ADMIN user:
image
Show graphQL error:
image

### DIFF
--- a/src/views/LogsView.js
+++ b/src/views/LogsView.js
@@ -1,0 +1,99 @@
+import React from 'react';
+import {graphql, compose} from 'react-apollo';
+import {ALL_EVENTS, MY_PROFILE} from '../state/queries';
+import {
+  Container,
+  Segment,
+  Message,
+  Placeholder,
+  Header,
+  Icon,
+  Select,
+} from 'semantic-ui-react';
+import EmptyView from './EmptyView';
+import EventList from '../components/EventList/EventList';
+import {eventType} from '../common/enums';
+
+const LogsView = ({events: {loading, allEvents, error, refetch}, user}) => {
+  const isAdmin = !user.loading
+    ? user.myProfile.roles.includes('ADMIN')
+    : false;
+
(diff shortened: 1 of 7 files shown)


src/views/LogsView.js
```diff
@@ -0,0 +1,125 @@
...
+              <p>User Error: {user.error.message}</p>
+            )}
+          </Message.Content>
+        </Message>
+      </Container>
+    );
+  if (isAdmin) {
+    return (
+      <Container as={Segment} basic vertical>
+        <Segment basic floated="right" className="noMargin noPadding">
```
dankolbman: Did you consider combining the event list with the existing component in the admin page?

src/views/LogsView.js
```diff
@@ -0,0 +1,125 @@
...
+              <Select
+                floated="right"
+                clearable
+                placeholder="Event Type"
+                options={eventTypeOptions}
+                onChange={(e, {name, value}) => setTypeFilter(value)}
+              />
+            </>
+          )}
+          <Button
```
dankolbman: Not sure if we really need a refresh button as maybe it's confusing with the other sync buttons.
IF we think that events are going to happen frequently, we could also consider enabling [polling](https://www.apollographql.com/docs/react/api/react-hoc/#datanetworkstatus).

bdolly:
@XuTheBunny would there be any downside to just hiding the `Logs` tab for non-ADMIN users?

src/views/LogsView.js
bdolly: I agree

Issue #393, opened 2019-08-06
Put download links behind a proxy path so that they can be real hyperlinks


fiendish:
This is related to https://github.com/kids-first/kf-ui-data-tracker/issues/191#issuecomment-518790388

The file download links use a dynamic javascript/graphql dance to generate a new single-use-only token for downloading every time you click the download button.

This negatively impacts the download links themselves because you can't right-click on them to do things like Copy URL or Save As.

Would it be possible to move that token and request dance one layer away through an `a` tag with href `"/download/study/{study}/file/{docid}/version/{verid}"` that's always available when the page is rendered?

dankolbman:
We may be able to make a 'sharable' url that presents a download of a specific version in the following way:
- Make versions have a static route in the app eg: `/studies/SD_XXX/documents/SF_XXX/version/FV_XXX`
- Have an additional route for `/download` relative to the previous route that will spawn the file download

This way, we can allow users to share that given route which, when clicked, will first auth the user through the regular flow, then point them to the version modal and immediately prompt the download.

The downside here is that the given route will still not be accessible through `curl` or standard GET requests to that endpoint.

fiendish:
> *The downside here is that the given route will still not be accessible through curl or standard GET requests to that endpoint.*

I don't follow why that needs to be true if the request can include authorization headers.

dankolbman:
Headers are not supported for standard GET requests within the browser.

fiendish:
You already proposed a mechanism for how to handle the browser.

dankolbman:
The `/studies/SD_XXX/documents/SF_XXX/version/FV_XXX` route provides a given version of a document in a study but may only be accessed with the correct `Authorization` header or download code. I'm not sure how we can utilize a proxy to turn these into real hyperlinks. I'm unsure of how to support some of the basic browser behavior while still solving our primary use cases.

fiendish:
When you click the current download link, where is the information that that uses to authenticate you stored?

PR #484, opened 2019-09-20
✨FileElement copy download link action button

### LINKED ISSUE #472: add button tooltips
Related to #399 probably.

Add tooltips to everything clickable, especially the ones that have consequences like the broken spatula and downvote buttons.
Screen Shot 2019-09-18 at 4 39 52 PM
Tiny icons are not unambiguous.

bdolly:
allow users to copy a download link the to most recent version of a document  #393

This copies `<download_url>/download/study/<study_id>/documents/<doc_id>/version/<version_id>` to the users clipboard
Screen Shot 2019-09-26 at 3 07 57 PM
Screen Shot 2019-09-26 at 3 07 53 PM

Closes #472

### DIFF
--- a/__mocks__/kf-api-study-creator/responses/studyByKfId.json
+++ b/__mocks__/kf-api-study-creator/responses/studyByKfId.json
@@ -32,6 +32,7 @@
                   {
                     "node": {
                       "id": "T2JqZWN0Tm9kZTpGVl80TlpOQkVPRg==",
+                      "kfId": "FV_NVV741M1",
                       "state": "PEN",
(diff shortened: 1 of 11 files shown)


bdolly:
> Should the kf_id copy button in the study header also be updated to keep iconography consistent?

Not at the moment. I think that button will likely get redesigned or go away in the near future as it's usefulness has been seen in usability testing

dankolbman:
Is it not trivial to update the icon? I think we also have another copy icon in the tokens page that should be updated.

PR #485, opened 2019-09-20
✨Feature document name validation

### LINKED ISSUE #217: Restrict file names to use standard conventions
File names upon upload and annotation should be validated for good practices as outlined in [this NIST document](https://www.nist.gov/sites/default/files/documents/pml/wmd/labmetrology/ElectronicFileOrganizationTips-2016-03.pdf).

bdolly:
## Enhancement
Adds text input checking/validation to the Document Form **Document Title** field using the following validation rules. 
Screen Shot 2019-09-20 at 4 39 29 PM

**Validation Rules**

| Rule       | RegEx           | 
| ------------- |:-------------:| 
| no words indicating a copied document    | new, final, modified, save(d), update(d), edit, ( 0-9 ) , [ 0-9 ]
| no special characters |  . $ @ & ! % * ] [ # ? / -
| no file extensions |  .[a-z]{2,}
| no specific dates |  mm-dd-yyy, mm-dd-yy, mm.dd.yyyy, mm.dd.yy, yyyymmdd, yymmdd 
| title must not be similar to file name |  >65% of Document Title bigrams must be different than those of the sanitized uploaded file name
| title must not be similar to existing document name in study |  >33% of Document Title bigrams must be different than those of existing documents in the study 

#### If any of the ☝️ rules fail the user will not be able to submit the EditDocumentForm
Screen Shot 2019-09-20 at 4 41 31 PM

#### If any similar documents are found it will suggest that the user "Update Existing Document"
Screen Shot 2019-09-20 at 4 44 18 PM

## Evidence 
As one of the goals of the Data Tracker application is to provide _reasonable_ constraints on user input it would be helpful to provide constraints such that they encourage users to input semantically meaningful **Document Title** 

During usability testing it was observed  ([testing observations] (https://airtable.com/tblkzD7aUN1a8Ne7e/viw1d4uUjKYQ89QMc/reciIQe80KtgtwGvk?blocks=hide)) that participants commonly named `new` study documents similar to the uploaded file name and when ask to upload a `new version ` of the same document they would create an entirely new document with the same name and append a qualifier such as `new, updated, modified` etc to the document name 

Closes #217


src/forms/EditDocument/ExistingDocsMessage.js
```diff
@@ -0,0 +1,73 @@
...
+          e.preventDefault();
+          setShowSimilar(!showSimilar);
+        }}
+      >
+        {' '}
+        {!showSimilar ? 'Show' : 'Hide'} Similar
+      </a>
+      {showSimilar && (
+        <List>
+          {similarStudyDocs.matches.map(doc => (
```
dankolbman: These don't render for me, even though the message does.

src/forms/EditDocument/validate.js
```diff
@@ -0,0 +1,74 @@
+/* eslint-disable no-useless-escape */
+import * as stringSimilarity from 'string-similarity';
+import {sortFilesBySimilarity} from '../../common/fileUtils';
+
+const MIN_SIMILARITY = 0.65;
```
dankolbman: Maybe move this to `src/common/globals.js`?

src/forms/EditDocument/validate.js
```diff
@@ -0,0 +1,74 @@
...
+    studyFiles,
+    0.33,
+  );
+
+  const uploadedFileSimilarity = stringSimilarity.compareTwoStrings(
+    DOC_NAME_INPUT,
+    cleanFileName,
+  );
+
+  // more than a quarter 75% of the inputed Document Title bigrams matches names of
```
dankolbman: I'm not sure what this comment means, but the code seems to imply that this is true if there is one or more docs that have a Dice similarity of 0.33 or more (twice the size of the intersecting bigrams by the sum of the size of the sets of bigrams).

src/modals/EditDocumentModal.js
```diff
@@ -11,63 +11,66 @@ const EditDocumentModal = ({
...
-        versionId: latestVersion.kfId,
-        description: latestVersion.description,
-        state: versionStatusInput,
-      },
-    })
-      .then(() => onCloseDialog())
-      .catch(err => console.log(err));
+  const handleSubmit = async (name, fileType, description, versionStatus) => {
+    try {
+      await updateFile({
```
dankolbman: Do you intend to do these synchronously, or can you use `Promise.all()`?

src/forms/EditDocument/ExistingDocsMessage.js
bdolly: they don't render when you click `Show Similar`?

src/forms/EditDocument/validate.js
bdolly: yeah I'll admit I don't understand the Dice similarity 💯 and was my attempt to explain

src/modals/EditDocumentModal.js
bdolly: this was a refactor of the existing function to attempt to catch all the errors in one place instead of individually. Wasn't sure if the order of operations mattered here when saving

src/forms/EditDocument/ExistingDocsMessage.js
```diff
@@ -0,0 +1,63 @@
...
+  );
+  const SimilarTitleContent = () => (
+    <>
+      We found &nbsp;
+      <strong>
+        {similarStudyDocs.matches.length}&nbsp; similar document titles &nbsp;
+      </strong>
+      in this study.
+      <details>
+        <sumamry>Show Similar</sumamry>
```
dankolbman: Typo here in `<summary>`

PR #489, opened 2019-09-24
✨ Edit Project modal

### LINKED ISSUE #464: Allow update of project fields
It would be helpful to be able to modify certain fields on projects such as:

- Workflow type
- Project type
- Name

For any fields that are shared with Cavatica, the changes should be propagated to Cavatica so that they are aligned in all places.

This could be done through an edit project modal.

dankolbman:
Adds a modal to edit existing Cavatica projects' workflow and project type.

Closes #464

Screen Shot 2019-09-24 at 11 14 39 AM


bdolly:
@dankolbman this needs to be refactored to include standard button color conventions used throughout the application and use the standard avatar and timestamp pattern

PR #491, opened 2019-09-25
✨ Feature UserSnap widget 

bdolly:
adds the UserSnap widget to the bottom right hand corner

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -25,6 +25,17 @@
       work correctly both with client-side routing and a non-root public URL.
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
+    <script type="text/javascript">
+      (function() {
+        var s = document.createElement('script');
+        s.type = 'text/javascript';
+        s.async = true;
+        s.src =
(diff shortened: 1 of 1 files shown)


public/index.html
```diff
@@ -25,6 +25,17 @@
       work correctly both with client-side routing and a non-root public URL.
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
+    <script type="text/javascript">
+      (function() {
+        var s = document.createElement('script');
+        s.type = 'text/javascript';
+        s.async = true;
+        s.src =
+          '//api.usersnap.com/load/64e9837d-529a-4f62-8322-350dca4c9f7c.js';
```
dankolbman: Missing a protocol?

bdolly: I can add one but that's copy/paste from their installation code. Maybe they detect protocol in their scripts?

dankolbman: Ah yeah I see it. Probably not handled by their script because it's in a script tag, so the browser will have to handle it. I guess it must be valid if it's in the  docs.

PR #495, opened 2019-10-01
✨ Add alert for study data completeness to study info page

XuTheBunny:
**Feature:**
- For ADMIN user, they can see the red corner label on missing value field on the study info page
- For ADMIN user, they can see the data completion progress bar on study info steps when editing 
- For ADMIN user, they can see the missing value field warning message on top of study info steps

**UI changes:**
Study basic info page:
Before:
On read-only mode:
image
On editing mode:
image
After:
On read-only mode:
image
image
image
On editing mode:
image
image
image
image

Study cavatica page:
Before:
image
After:
Screen Shot 2019-10-01 at 3 13 47 PM


src/App.css
```diff
@@ -154,3 +154,16 @@ body {
 .text-wrap-75 {
   max-width: 32em;
 }
+
+.full-width {
+  /* width: 70%; */
+}
```
bdolly: 🔥

src/App.css
```diff
@@ -154,3 +154,16 @@ body {
...
+
+.progress-bar {
```
bdolly: this may need to be more specific so it doesn't affect other general progress bars later

src/components/StudyHeader/StudyHeader.js
```diff
@@ -26,9 +26,11 @@ const StudyHeader = ({
   return (
     <Container>
       <Header as="h1">{studyName}</Header>
-      <Header.Subheader as="h2" className="study--header-sub">
-        {shortName}
-      </Header.Subheader>
+      {shortName && (
+        <Header.Subheader as="h2" className="study--header-sub">
```
bdolly: why the class?

src/components/StudyInfo/ProgressBar.js
```diff
@@ -0,0 +1,32 @@
...
+      total={total}
+      size="tiny"
+      label={filled + '/' + total + ' fields complete'}
+      error={filled === 0}
+      warning={filled > 0 && filled < total}
+      success={filled === total}
+    />
+  );
+};
+
```
bdolly: propTypes will help understand this component a bit more

src/forms/FormField.js
```diff
@@ -22,44 +23,52 @@ const FormField = ({
...
+          className={readOnly ? 'readOnlyField noMargin' : 'noMargin'}
+          fluid
+          readOnly={readOnly}
+          type={type}
+          name={id}
+          placeholder={placeholder}
+          open={focused}
+          onChange={handleChange}
+          onBlur={e => {
+            handleFocus('');
```
bdolly: why and empty string param? could it be `null` instead?

src/forms/FormField.js
```diff
@@ -22,44 +23,52 @@ const FormField = ({
...
-      <Label pointing basic color="red">
-        {errors}
-      </Label>
-    )}
-  </Form.Field>
-);
+  isAdmin,
+}) => {
+  const hasError = touched && errors && errors.length > 0;
+  const tracking =
```
bdolly: not sure what this does 🤷‍♂

src/forms/NewStudyForm.js
```diff
@@ -45,6 +56,33 @@ const NewStudyForm = ({
       href: 'logistics',
     },
   ];
+  const missingValueMessage = values => {
+    var fields = [];
+    trackedStudyFields.map(field => {
+      if (!values[field] || values[field].length === 0 || values[field] === 0) {
+        fields.push(fieldLabel[field]);
```
bdolly: is there a way to map what field goes to what step so we can link them to the step the field is in?

src/forms/StudyFormSteps.js
```diff
@@ -38,6 +38,7 @@ export const InfoStep = ({
   newStudy,
```
bdolly: 🚚 let's modularize this form a bit into a directory 
````
/forms 
|__/StudyInfoForm
   |__ /Steps
   |   |__ InfoStep.js
   |   |__ LogisticsStep.js
   |   |__ ExternalProjectStep.js
   |   |__ index.js
   |__ index.js (form component)  
```

src/forms/StudyFormSteps.js
```diff
@@ -48,6 +49,7 @@ export const InfoStep = ({
         content="Please provide the study's full name and a shortened version that may be used for display purposes."
       />
       <FormField
```
bdolly: alot of the fields in this form look very similar let's see if we can map them instead of repeated them to make each step file a little smaller

src/components/StudyHeader/StudyHeader.js
XuTheBunny: There was some extra space before the copy study id button when there is no short name. This will remove those space

src/forms/FormField.js
XuTheBunny: This will return true or false for us to show/hide the red corner label on each field. We want to check if the current user is ADMIN, if the field is one of the fields we want to track, and if it has already had value. I put it outside the component instead of inline to make the component looks cleaner.

src/forms/StudyInfoForm/Steps/FormField.js
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
```
bdolly: can this be abstracted/generalized enough to move to `src/forms/FormField.js` ?

src/forms/StudyInfoForm/NewStudyForm.js
```diff
@@ -45,6 +63,52 @@ const NewStudyForm = ({
...
+        flatList.indexOf(field) >=
+        steppingFields[0].length + steppingFields[1].length
+      ) {
+        stepIndex = 2;
+      }
+      if (!values[field] || values[field].length === 0 || values[field] === 0) {
+        fields.push({label: fieldLabel[field], step: stepIndex});
+      }
+      return fields;
+    });
```
bdolly: could possibly shorten this slightly to 
```js
let fields = steppingFields.reduce((acc, stepsArr, step) => {
      return acc.concat(
        stepsArr
          .filter(
            field =>
              trackedStudyFields.includes(field) &&
              (!values[field] ||
                values[field].length === 0 ||
                values[field] === 0),
          )
          .map(f => ({label: fieldLabel[f], step})),
      );
    }, []);
```

PR #524, opened 2019-10-22
📈Study Grid view events tracking

bdolly:
## Feature or Improvement
Tracks events for `StudyListView` these include: 
#### STUDY_LIST
-  **STUDY_LIST__VIEW_TOGGLE**  view toggle button 
- **STUDY_LIST__SEARCH** "Search by Study Name" search input results length
- **STUDY_LIST__ADD_STUDY** "Add Study" button events 

#### STUDY_CARD
- **STUDY_CARD__TOGGLE_DETAIL**
- **STUDY_CARD__TOOLTIP_FILES__CLICK**
- **STUDY_CARD__TOOLTIP_INFO__CLICK**
- **STUDY_CARD__TOOLTIP_PROJECTS__CLICK**

Event schemas can be found in src/analyticsTracking/event_schemas

Adds the `wtihAnalyticsTracking` HOC that passes down the Amplitude component props as well as a `tracking` prop object that provides convenience methods for event tracking.

## UI changes
n/a 

| Browser             | 
| ------------------- |
|  ✅ Chrome (77.0.3865.90 )  |    
| ✅ Firefox (69.0.1) |  
| ✅Safari (12.1.2)  |     
| ✅  Edge 42   |


src/components/StudyInfo/CavaticaCounts.js
```diff
@@ -19,7 +25,21 @@ const CavaticaCounts = ({projects, title, hideIcon}) => {
       <List.Item
         as={Link}
         to={`/study/${title}/cavatica`}
-        onClick={e => e.stopPropagation()}
+        onClick={e => {
+          e.stopPropagation();
+          logEvent(
+            inheritedEventProps.scope
+              ? inheritedEventProps.scope + '__TOOLTIP_PROJECTS__CLICK'
```
dankolbman: Why not use an array to describe the scope like [this](https://github.com/amplitude/react-amplitude#example-instrumenting-tic-tac-toe-from-facebooks-intro-to-react-tutorial)?

src/components/StudyInfo/CavaticaCounts.js
```diff
@@ -65,4 +85,4 @@ CavaticaCounts.defaultProps = {
   projects: [],
 };
 
-export default CavaticaCounts;
+export default withAnalyticsTracking(CavaticaCounts, {logToConsole: true});
```
dankolbman: Why do you need this? Doesn't the amplitude provider pass down all the inherited info you need?

src/components/StudyInfo/CavaticaCounts.js
bdolly: thought about that. I think we get the same level of information by having a `view` prop in every event that is the name of the current view from the route then `scope` just applies to the specific container in the view. I think this also makes it easier to query from the Amplitude dashboard(s) as you don't have to check for something in an array. 

If we do use scope as an array it would be a better candidate to roll into the AmplitudeProxy to pass with every call

src/components/StudyInfo/CavaticaCounts.js
bdolly: because I think render props are messy and this allows you to add `eventProperties` prop directly to the component. It also infuses the renderProps in AmplitudeProxy with the EVENT_CONSTANTS and allows use to hook into other dev. exp. conveniences like logging events to the console and saving event json schemas to a local file when building out new new events

package.json
```diff
@@ -68,7 +68,10 @@
     "@babel/plugin-transform-runtime": "^7.4.4",
     "@babel/preset-env": "^7.4.3",
     "@babel/preset-react": "^7.0.0",
+    "@isomorphic-git/lightning-fs": "^3.4.1",
```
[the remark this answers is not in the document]
bdolly: 💩 sloppiness on my part, an artifact left over from experimentation

PR #548, opened 2019-11-25
✨ Add study-creator-api version to footer

### LINKED ISSUE #544: Add API version
## Persona

Users

## Motivation/Goal

Add the version number of the study creator that results from the `status` query to the footer of the Data Tracker so that we may more easily determine currently deployed version.

## Supporting Evidence (qual or quant)

It is currently not possible to tell what version of the Study Creator API the Data Tracker is using when a user interacts with the application.

XuTheBunny:
## Feature or Improvement

Add the version number of the study creator that results from the `status` query to the footer of the Data Tracker so that we may more easily determine currently deployed version.

## UI changes

**The footer on any page:**
**Before:**
Screen Shot 2019-11-25 at 3 11 03 PM
**After:**
Screen Shot 2019-11-25 at 3 10 52 PM

## Browsers Tested

| Browser             | 1024px | 768px | 480px | 320px |
| ------------------- | :----: | ----: | ----: | ----: |
| Chrome (77)  | ✅ | ✅ | ✅ | ✅ |
| Firefox (69) | ✅ | ✅ | ✅ | ✅ |
| Safari (12)      |  ✅ | ✅ | ✅ | ✅ |
| Edge (18)      |   ✅ | ✅ | ✅ | ✅ |

Closes #544

### DIFF
--- a/__mocks__/kf-api-study-creator/mocks.js
+++ b/__mocks__/kf-api-study-creator/mocks.js
@@ -6,6 +6,7 @@ import {
   ALL_EVENTS,
   ALL_USERS,
   GET_PROJECTS,
+  STATUS,
 } from '../../src/state/queries';
 import {
   CREATE_PROJECT,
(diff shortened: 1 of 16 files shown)


bdolly:
@XuTheBunny I think since the information presented here is esoteric we could just have it say `API` and maybe drop the font-size a bit

PR #627, opened 2020-03-24
📈 File actions events

dankolbman:
## Feature or Improvement

Adds events to the file actions on both the file list view and the file detail view.


dankolbman:
Needs to be updated after being refactored.

PR #701, opened 2020-05-11
✨ Persist study list column sort state

### LINKED ISSUE #694: Persist study list column sort state
## Requirements

The column and direction of the study list sorting should be saved and restored from localStorage similar to how the visible columns are.

Screenshot_2020-05-05 KF Data Tracker - My Studies(1)

## Motivation/Goal

Users may often be interested in the same order of studies based on their desired criteria so that they may step through each individually. Having to resort the table upon returning to the study list is time consuming and not effective.

XuTheBunny:
Save study table sorting state to local storage (`sortingColumn` and `sortingDirection`)

Closes #694

### DIFF
--- a/src/components/StudyList/StudyList.js
+++ b/src/components/StudyList/StudyList.js
@@ -302,7 +302,8 @@ const StudyList = ({studyList, loading, activeView, history, myProfile}) => {
                   myProfile={myProfile}
                   loading={loading}
                   studyList={filteredStudyList()}
-                  columns={columns.columns}
+                  columns={columns}
+                  handleSort={handleSort}
                 />
               )}
             </Grid.Column>
--- a/src/components/StudyList/StudyList.js
+++ b/src/components/StudyList/StudyList.js
@@ -98,6 +98,27 @@ const StudyList = ({studyList, loading, activeView, history, myProfile}) => {
         },
   );
 
+  const handleSort = column => () => {
+    const direction =
+      columns.sorting.column !== column
+        ? 'ascending'
+        : columns.sorting.direction === 'ascending'
+        ? 'descending'
+        : 'ascending';
+    setColumns({
+      ...columns,
+      sorting: {column, direction},
+    });
+    localStorage.setItem(
+      'studyColumns',
+      JSON.stringify({
+        columns: existingState ? existingState.columns : defaultState.columns,
+        version: defaultState.version,
+        sorting: {column, direction},
+      }),
+    );
+  };
+
   if (loading) {
     return (
       <Container as={Segment} basic>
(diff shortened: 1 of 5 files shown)


src/components/StudyList/StudyTable.js
```diff
@@ -61,27 +64,33 @@ const StudyTable = ({
   columns,
 }) => {
   const [sorting, setSorting] = useState({
-    column: 'name',
-    direction: 'descending',
+    column:
```
dankolbman, line 67: Could we move this up into the `StudyList` component along with the rest of the column state:
https://github.com/kids-first/kf-ui-data-tracker/blob/master/src/components/StudyList/StudyList.js#L68-L99
That will make the one column state variable the single key in localStorage that deals with all the table display values.

src/components/StudyList/StudyList.js
```diff
@@ -98,6 +98,32 @@ const StudyList = ({studyList, loading, activeView, history, myProfile}) => {
         },
   );
 
+  const [sorting, setSorting] = useState({
```
dankolbman, line 101: Could you move this into `columns` so it looks like:
```
{
  version: 3,
  columns: [ ... ],
  sorting: { column: '', direction: ''}
}
```

src/components/StudyList/StudyList.js
```diff
@@ -65,38 +65,66 @@ const StudyList = ({studyList, loading, activeView, history, myProfile}) => {
...
-            },
-            {
-              key: 'phenotypeStatus',
-              name: 'Phenotype Status',
-              visible: false,
-            },
-            {key: 'version', name: 'Version', visible: true},
-          ],
-        },
+    existingState && existingState.version === 2 ? existingState : defaultState,
```
dankolbman, line 101: This needs to check that it is at version 3 or it will remove the state everytime.

src/components/StudyList/StudyList.js
```diff
@@ -65,38 +65,66 @@ const StudyList = ({studyList, loading, activeView, history, myProfile}) => {
...
+  );
+  const [sorting, setSorting] = useState(
```
dankolbman, line 103: You shouldn't need this because it's in the `columns` variable now.

src/components/StudyList/StudyList.js
dankolbman: maybe check `existingState.version === defaultState.version` here so we don't need to update it everytime the default state schema changes.

src/components/StudyList/StudyList.js
```diff
@@ -65,39 +65,63 @@ const StudyList = ({studyList, loading, activeView, history, myProfile}) => {
...
+        : 'ascending';
+    setColumns({
+      ...columns,
+      sorting: {column, direction},
+    });
+    localStorage.setItem(
+      'studyColumns',
+      JSON.stringify({
+        columns: existingState ? existingState.columns : defaultState.columns,
+        version: 2,
```
dankolbman, line 119: we should always be able to use `defaultState.version` here.

PR #738, opened 2020-06-09
✨ Add preview for user group permissions in users screen

### LINKED ISSUE #719: Admin screen for viewing user group permissions
## Requirements

Admins will want to know which permissions each user group has. We should build a permissions view that displays permissions grouped by permission group, perhaps as part of the users admin view.

XuTheBunny:
Show permission group in admin user view
Screen Shot 2020-06-10 at 8 28 59 PM

Show permission group in invite user modal
Screen Shot 2020-06-10 at 10 59 51 PM
Screen Shot 2020-06-10 at 11 00 10 PM

Closes #719


src/admin/components/UserList/PermissionGroup.js
```diff
@@ -0,0 +1,136 @@
...
+              onClick={titleProps =>
+                setActiveIndex(activeIndex === group.key ? '' : group.key)
+              }
+              data-testid="user-group-header"
+            >
+              <Icon name="dropdown" />
+              {group.key}
+            </Accordion.Title>
+            <Accordion.Content active={activeIndex === group.key}>
+              {group.key === 'Administrators' ? (
```
dankolbman, line 28: Instead of only showing multiple columns for only the admin group, why not just always split it into 3 columns?

src/admin/components/UserList/PermissionGroup.js
```diff
@@ -0,0 +1,136 @@
...
+                <Grid columns={3}>
+                  <Grid.Column>
```
dankolbman, line 30: Instead of repeating `<Grid.Column>` the same way each time, why not make a component that takes a list of permissions and returns the column so it doesn't need to be repeated?

PR #753, opened 2020-06-13
✨ Add missing fields to configuration page 

### LINKED ISSUE #742: Add missing fields to configuration page
## Requirements

The configuration page should list all settings and feature flags.
These fields should be able to be dynamically updated from the api so we don't need to update the status query every time it changes in the backend.

These queries should return all fields on the settings and features:
```
query features {
  __type(name: "Features") {
    name
    fields {
      name
    }
  }
}

query settings {
  __type(name: "Settings") {
    name
    fields {
      name
    }
  }
}
```

XuTheBunny:
Call the following queries to get the fields for Status query:
```
query features {
  __type(name: "Features") {
    name
    fields {
      name
    }
  }
}

query settings {
  __type(name: "Settings") {
    name
    fields {
      name
    }
  }
}
```

Status query:
```
query Status {
  status {
    name
    version
    commit
    features {
      ...FeaturesFields
    }
    settings {
      ...SettingsFields
    }
    queues
    jobs {
      edges {
        node {
          id
          name
          active
          failing
          lastRun
          lastError
          createdOn
          enqueuedAt
        }
      }
    }
  }
}
${FEATURES_FIELDS}
${SETTINGS_FIELDS}
```

Closes #742

### DIFF
--- a/src/admin/views/ConfigurationView.js
+++ b/src/admin/views/ConfigurationView.js
@@ -10,7 +10,8 @@ import {
   Dimmer,
 } from 'semantic-ui-react';
 import TimeAgo from 'react-timeago';
-import {STATUS} from '../../state/queries';
+import {FEATURES, SETTINGS} from '../../state/queries';
+import gql from 'graphql-tag';
 
 const FeatureTable = ({features}) => (
   <Table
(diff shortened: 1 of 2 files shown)


src/admin/views/ConfigurationView.js
```diff
@@ -147,6 +148,119 @@ const LoadingTable = () => (
 );
 
 const ConfigurationView = () => {
+  var STATUS = gql`
```
dankolbman, line 151: Do you need to have this definition of the status query? Instead of relying on refetchQueries, maybe  the status query can be inside a container component and rendered after the fragments are built so we don't have to make many different calls to the status query.

PR #754, opened 2020-06-15
✨ Include referrals in tracked routes

### LINKED ISSUE #752: Include referrals in tracked routes
If a page a user is viewing has a `utm_source` in the query parameters, we should include that in the event properties. If there is no `utm_source` we don't need to add it.

https://github.com/kids-first/kf-ui-data-tracker/blob/3aa703bfab225d54c123356be2c4c8ed63bd9541/src/routes/TrackedRoute.js#L19-L26

XuTheBunny:
Only add `utm_source` prop to Amplitude event properties when url contains `?utm_source=XXX`
URL: `http://localhost:3000/study/SD_W2PQV9FJ/logs?utm_source=slack`
Screen Shot 2020-06-15 at 4 42 33 PM
URL: `http://localhost:3000/study/SD_W2PQV9FJ/releases`
Screen Shot 2020-06-15 at 4 42 42 PM

Closes #752

### DIFF
--- a/src/routes/TrackedRoute.js
+++ b/src/routes/TrackedRoute.js
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import {Route} from 'react-router-dom';
 import {Amplitude} from '@amplitude/react-amplitude';
+import queryString from 'query-string';
 
 /**
  * Wrap all of our routes to make them log the view and route they render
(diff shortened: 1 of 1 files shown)


src/routes/TrackedRoute.js
```diff
@@ -15,27 +16,40 @@ const TrackedRoute = ({
...
-          <Route component={component} {...rest} />
-        )}
-        {!disableViewEvent && logEvent('page view')}
-      </>
-    )}
-  </Amplitude>
-);
+}) => {
+  const eventProps = inheritedProps => {
+    if (rest.location.search) {
```
dankolbman, line 21: You could probably save yourself from needing to repeat the full object within the `if` by just adding it to an existing object conditionally:
```
var eventProps =  { ...inheritedProps, ... };

if (rest.location.search) eventProps = { ...eventProps, utm_source: ... };
```

PR #772, opened 2020-06-23
💄 Render different document help message on file list and detail

### LINKED ISSUE #770: Split document "About" box into two
## Feature Justification

The same document about box shows on both the study document overview page and also on the document versions page, but only half of the box applies to either, so the box is box bigger up front and also only partly applicable to each.

Screen Shot 2020-06-19 at 8 16 29 AM

Split the box into two halves. On the study overview page (e.g. https://kf-ui-data-tracker.kidsfirstdrc.org/study/SD_ME0WME0W/documents) show:

> About Study Documents
>
> Many documents need to be collected in order to properly ingest a study into the Kids First Data Resource Center (DRC). 
> Required documents may be uploaded to a study here so that they may be reviewed and processed by the Kids First DRC.

and on the document versions page (e.g. https://kf-ui-data-tracker.kidsfirstdrc.org/study/SD_ME0WME0W/documents/SF_S8TW02HS) show

> About Document Versions
>
> Versions allow revisions of the same document to be tracked. Creation of a new document that contains revised data of 
> another document should be avoided and instead uploaded as a new version of the existing document.

XuTheBunny:
Screen Shot 2020-06-23 at 2 54 45 PM
Screen Shot 2020-06-23 at 3 27 15 PM

Closes #770


src/documents/components/helpers/DocumentListHelp.js
```diff
@@ -2,20 +2,23 @@ import React from 'react';
...
-      uploaded to a study here so that they may be reviewed and processed by the
-      Kids First DRC.
-    </p>
-    <p>
-      <b>Versions</b> allow revisions of the same document to be tracked.
-      Creation of a new document that contains revised data of another document
-      should be avoided and instead uploaded as a new version of the existing
-      document.
-    </p>
+    {history.location.pathname.includes('SF_') ? (
```
dankolbman, line 7: Can you instead split these into different helpers and render them via the study router?

PR #819, opened 2020-07-30
✨ Add checkbox in filter bar to show hide file ID column on file table

XuTheBunny:
Screen Shot 2020-07-30 at 2 56 01 PM


src/documents/components/FileList/FileList.js
```diff
@@ -134,6 +135,16 @@ const FileList = ({
               >
                 Document Details
               </Table.HeaderCell>
+              {showId && (
+                <Table.HeaderCell
+                  sorted={sorting.column === 'kfId' ? sorting.direction : null}
```
dankolbman, line 140: `kfId` may need to be added to the `columnSorts` so it sorts correctly.

PR #861, opened 2020-09-08
✨ Add document extract config modal to preview and download config file 

XuTheBunny:
## Visual Changes
image
Screen Shot 2020-09-10 at 1 37 19 PM
Screen Shot 2020-09-10 at 1 37 01 PM

### DIFF
--- a/src/documents/components/FileDetail/FileDetail.js
+++ b/src/documents/components/FileDetail/FileDetail.js
@@ -92,6 +92,7 @@ const ActionButtons = ({
             }
             color="grey"
             icon="download"
+            className="mb-15"
             fluid
             size="mini"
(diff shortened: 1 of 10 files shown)


src/documents/components/FileDetail/FileDetail.js
```diff
@@ -103,6 +104,22 @@ const ActionButtons = ({
           />
         }
       />
+      <Popup
```
dankolbman, line 107: Not all file types will have an extract config, so we need to perhaps add a key in the `fileTypeDetail` enum for that and then only display this button for those valid file types.

dankolbman (approved):
💯

PR #862, opened 2020-09-09
⚡ Pare down fields requested for all studies

dankolbman:
Reduces the amount of information requested to retrieve all studies to display on the study list view.


XuTheBunny (approved):
🌰

PR #877, opened 2020-09-17
✨ Disable the file type selection option if file is not qualified

XuTheBunny:
Screen Shot 2020-09-17 at 12 14 37 PM


dankolbman (commented):
Do you think it's better to hide the invalid types or just disable them?

dankolbman (approved):
:100:

PR #882, opened 2020-09-22
🔥 Remove obsolete new document view and /new-document route

XuTheBunny:
`/new-document` route is not longer in use, we only use `/upload`


dankolbman (approved):
:100:

PR #893, opened 2020-09-24
🔧 Redirect host

dankolbman:
Redirects the application to the desired host if given. This allows us to funnel multiple deployments at different domains to one central location.


XuTheBunny (approved):
🥧

PR #897, opened 2020-09-30
🐛 Fix user group dropdown in user page not updating

### LINKED ISSUE #896: User group dropdown in user page is not responding on selection
image

The data wasn't updated on selection, lacking the refetch query.

XuTheBunny:
Closes #896


dankolbman (approved):
🥮

PR #900, opened 2020-10-08
✨ Add copy download url button and copy version kf ID button

### LINKED ISSUE #895: Add copy to clipboard buttons to copy file and version download link
It will be convenient that we can have a button to copy the download link to clipboard. Instead of right click on the download button and try to get the link from there.

XuTheBunny:
Screen Shot 2020-10-07 at 9 13 48 PM
Screen Shot 2020-10-07 at 9 14 06 PM
Screen Shot 2020-10-07 at 9 14 19 PM

Closes #895

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -191,6 +191,14 @@ body {
   color: rgba(0, 0, 0, 0.7);
 }
 
+.smallButton {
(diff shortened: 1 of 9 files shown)


dankolbman (approved):
🖲

PR #901, opened 2020-10-08
♻️ Default showing only my studies to be false

XuTheBunny:
image

- For all new users, default to display all the studies.
- For users who cleared their local storage (not having `onlyMyStudies` value), default to display all the studies.
- For user who have `onlyMyStudies` value, toggling the only showing my studies based on its value.


dankolbman (approved):
👍

PR #902, opened 2020-10-08
🔧 Add dev/QA amplitude key to netlify builds

dankolbman:
Adds an Amplitude key for our dev/QA data source to the qa and dev builds on netlify. The production amplitude key will get inherited from the Netlify UI on production builds.


XuTheBunny (approved):
🥟

PR #903, opened 2020-10-09
📝 Update setup steps in README

znatty22:
I went through the README to try and setup the Data Tracker dev env and came across some things. Here's an updated README with some fixes/clarifications. Not sure if all are valid (do we still use kf-data-stack??)

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,34 +13,34 @@ Investigator study management and progress tracking.
 
 ## Development Quick Start
 
+If you haven't already, checkout the kf-ui-data-tracker master branch:
+
+```
+git clone https://github.com/kids-first/kf-ui-data-tracker.git
+cd kf-ui-data-tracker
+```
+
 #### Setup backend API service
 
 Study-creator served as backend provides the data APIs consumed by data-tracker frontend.
 
-If you haven't already, create the kf-data-stack network:
+If you haven't already, create the kf-data-stack network and necessary external volume:
 
 ```
 docker network create kf-data-stack
+docker volume create --name=bundle
 ```
 
-Checkout kf-api-study-creator master and bring up the study creator API:
-Checkout the [Kids First Data Stack](https://github.com/kids-first/kf-data-stack)
-and make sure the study creator services are running:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -13,35 +13,35 @@ Investigator study management and progress tracking.
...
 
 ```
 docker network create kf-data-stack
+docker volume create --name=bundle
 ```
 
-Checkout kf-api-study-creator master and bring up the study creator API:
-Checkout the [Kids First Data Stack](https://github.com/kids-first/kf-data-stack)
-and make sure the study creator services are running:
+Next bring up the study creator API in development mode:
```
znatty22, line 34: Not really sure what was supposed to go here. Do we use kf-data-stack docker-compose, study-creator docker-compose or data-tracker docker-compose?

dankolbman: The docker-compose included in this repo sets up the study creator and all other apis needed for testing. The line you have should work for just bringing up the essential study creator api.

README.md
```diff
@@ -64,6 +64,9 @@ REACT_APP_AUTH0_AUD
 
 See the provided `.env.schema` file for more details.
 
+You may need to contact the project maintainer to get the appropriate value for
```
dankolbman, line 67: We can replace with with 
> You will need to setup an application in Auth0 and use it to set the `REACT_APP_AUTH0_CLIENT_ID` variable. Make sure to configure the application to allow redirects at the local server's address (typically http://localhost:3000/callback)

znatty22: Ok so is it expected that each developer will need to setup an application in Auth0 to get their client ID? Or is this blurb for the general developer outside Kids First? In other words, are all Kids First developers expected to use the same client id?

dankolbman: Devs outside Kids First should need to set up their own Auth0 app. Devs in Kids First could use the same application client id or set them up individually.

dankolbman (approved):
# 🍄

Thanks, we definitely need to do a lot of cleanup with the developer side of things from a better readme to just organizing our config better.

PR #909, opened 2020-10-21
🔥  Remove status fields from study table

### LINKED ISSUE #880: Remove status fields from study table.
## Feature Justification

The status fields do not tend to be used for studies in the Data Tracker. We should remove these columns from the study table.

Because we will be updating the allowed columns in the study table, we will have to do an in place migration so that anyone who previously had these columns enabled won't have their app break when they try to log in next time.

Screenshot_2020-09-21 KF Data Tracker - My Studies

XuTheBunny:
image

Closes #880

### DIFF
--- a/cypress/integration/studies/study_statuses.spec.js
+++ b/cypress/integration/studies/study_statuses.spec.js
@@ -1,5 +1,9 @@
 /// <reference types="cypress" />
 
+/*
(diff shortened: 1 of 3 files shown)


dankolbman (approved):
🧞‍♂️

PR #911, opened 2020-10-26
✨ Add genomic workflow output manifest to file types

znatty22:
Adds a new expedited file type for the Genomic Workflow Output Manifest to support https://github.com/kids-first/kf-api-study-creator/issues/492

## Visual Changes
image


dankolbman (approved):
# 🥬

PR #912, opened 2020-10-27
✨ Change releases to study creator backend

dankolbman:
This removes any calls to the Release Coordinator api and replaces them with calls to the Study Creator, which now supports releases.


dankolbman:
Yeah, our compose file here can change now that we don't need to stand up coordinator containers.

XuTheBunny (approved):
🍣

PR #913, opened 2020-10-27
👷‍♂️ Use netlify production as qa

dankolbman:
This sets the netlify production environment to redirect to the qa url instead of the master. This will allow us to access qa api and interface through https://kf-ui-data-tracker-qa.kidsfirstdrc.org.


XuTheBunny (approved):
🌮

PR #932, opened 2020-11-30
⚡️ Make e2e tests less awful

dankolbman:
This should avoid having to download and install cypress for every CI integration test run.


dankolbman:
and it didn't improve test times :(

PR #933, opened 2020-12-01
🔧 Add developer views

dankolbman:
Adds a schema explorer and GraphiQL to help developers.

Current user's token is automatically attached to all requests.

### GraphiQL
Screen Shot 2020-12-01 at 12 03 42

### Voyager
Screen Shot 2020-12-01 at 10 39 10


XuTheBunny (approved):
🍷

PR #948, opened 2020-12-11
💄 Increase hit area for releases

dankolbman:
Makes the entire cell clickable to navigate to a specific release and makes the table selectable.

## Visual Changes

Screen Shot 2020-12-11 at 4 22 59 PM


XuTheBunny (approved):
🍩

PR #957, opened 2021-01-07
✨ App banners

### LINKED ISSUE #958: App banners
## Feature Justification

Administrators may want to add message banners that can be persistently displayed on Data Tracker, informing users of anything important (e.g. scheduled downtime, new feature announcement).

Add new admin view for listing, adding, editing, deleting banners

znatty22:
Closes #958 

## Description
A new feature to create, update, delete, and edit message banners that will be persistently displayed on Data Tracker when we want to notify users of something important (e.g. maintenance, new features, etc)

## Visual Changes

image

image

image

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -2,12 +2,12 @@
 @import '../node_modules/react-datasheet/lib/react-datasheet.css';
 
 :root {
-  /* define global css vars here  */
-  --site-max-width: 1024px;
+    /* define global css vars here  */
+    --site-max-width: 1024px;
 }
 
 * {
-  box-sizing: border-box;
+    box-sizing: border-box;
 }
 
 body {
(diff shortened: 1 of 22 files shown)


src/admin/components/BannerTable/ActionButtons.js
```diff
@@ -0,0 +1,60 @@
...
+import {
+  Icon,
+  Divider,
+  Responsive,
+  Popup,
+  Button,
+  Table,
+} from 'semantic-ui-react';
+
+const ActionButtons = ({handleDelete, handleEdit}) => {
```
dankolbman, line 11: Since you don't have any state for this component, you can return the JSX as a functional component:

```js
const ActionButtons = ({handleDelete, handleEdit}) => (
    <Table.Cell>
        ...
    </Table.Cell>
)
```

src/admin/components/BannerTable/BannerTable.js
```diff
@@ -0,0 +1,90 @@
...
+const COLUMNS = [
+  {key: 'message', name: 'Message'},
+  {key: 'creator', name: 'Creator'},
+  {key: 'severity', name: 'Severity'},
+  {key: 'startDate', name: 'Start'},
+  {key: 'endDate', name: 'End'},
+  {key: 'enabled', name: 'Enabled'},
+];
+
+function BannerTable({loading, error, banners, handleDelete, handleEdit}) {
```
dankolbman, line 19: We tend to declare all components as arrow functions:
```js
const BannerTable = ({props}) => {
    ...
}
```

src/admin/forms/BannerForm.js
```diff
@@ -0,0 +1,217 @@
...
+  {key: 'ERROR', name: 'Error'},
+];
+
+const BannerForm = ({history, isUpdate, mutation, banner, loading, err}) => {
+  const [submitting, setSubmitting] = useState();
+  const [error, setError] = useState(err);
+  const [state, setState] = useState({
+    message: banner.message,
+    enabled: banner.enabled,
+    startDate: banner.startDate ? banner.startDate.split('T')[0] : '',
```
dankolbman, line 17: I think you should be able to interpret these ISO8601 dates with `new Date(banner.startDate)` and use `toLocaleString` for formatting.

src/admin/forms/BannerForm.js
```diff
@@ -0,0 +1,217 @@
...
+const SEVERITY = [
+  {key: 'INFO', name: 'Info'},
+  {key: 'WARN', name: 'Warning'},
+  {key: 'ERROR', name: 'Error'},
+];
+
+const BannerForm = ({history, isUpdate, mutation, banner, loading, err}) => {
+  const [submitting, setSubmitting] = useState();
+  const [error, setError] = useState(err);
+  const [state, setState] = useState({
```
dankolbman, line 14: This looks like it will work, although we tend to use `Formik` to set up all and store all the state, submission, and validation for us. Perhaps you want to try wrapping your form with it to see how it works?

src/admin/views/BannersView.js
```diff
@@ -0,0 +1,92 @@
...
+      )}
+      <Header as="h3">Banners</Header>
+
+      {/* Banners Table */}
+      {!error && (
+        <BannerTable
+          error={error}
+          loading={loading}
+          banners={data && data.allBanners.edges}
+          handleDelete={handleDelete}
```
dankolbman, line 84: One trick we use is to check if the user has permission to do an action here in a container component and if they don't, we pass in `null` for the function. Then, in the child component, we hide the restricted actions if the function is null.

src/admin/views/EditBannerView.js
```diff
@@ -0,0 +1,69 @@
...
+  Container,
+  Segment,
+  Header,
+  Message,
+  Dimmer,
+  Loader,
+} from 'semantic-ui-react';
+import {UPDATE_BANNER} from '../mutations';
+import {GET_BANNER, ALL_BANNERS} from '../queries';
+import WrappedBannerForm from '../forms/BannerForm';
```
dankolbman, line 14: Why call this `WrappedBannerForm` and not just `BannerForm`?

src/admin/queries.js
```diff
@@ -149,3 +149,47 @@ export const ALL_JOBS = gql`
     }
   }
 `;
+
+// Get all banners
+export const ALL_BANNERS = gql`
+  query Banners {
```
dankolbman, line 155: I think we can filter this by `enabled: true` to reduce what's being sent over the network?

src/admin/views/EditBannerView.js
znatty22: Haha I saw it somewhere else in the code and thought it was some sort of convention.

src/admin/queries.js
dankolbman: Yeah, I was suggesting that you add a variable to the query to allow you to specify if you want enabled banners or not:
```
query Banners($enabled: Boolean) {
    ...
}
```

src/admin/components/BannerTable/BannerTable.js
```diff
@@ -0,0 +1,100 @@
...
+            <BannerMessage key={node.id + 'message'} message={node.message} />
+            <Creator
+              key={node.id + 'creator'}
+              displayName={node.creator.displayName}
+              picture={node.creator.picture}
+            />
+            <Severity key={node.id + 'severity'} severity={node.severity} />
+            <BannerDate key={node.id + 'startDate'} date={node.startDate} />
+            <BannerDate key={node.id + 'endDate'} date={node.endDate} />
+            <Table.Cell singleLine>
```
dankolbman, line 75: It would be handy if this was a toggle button so we could immediately change the enabled/disabled state without going to the edit page.

src/components/Header/Header.js
```diff
@@ -153,60 +160,63 @@ const Header = ({location}) => {
...
-              <Menu.Item as={Nav} to="/releases/history">
-                Releases
-                <Label content="beta" color="blue" attached="bottom right" />
-              </Menu.Item>
-            )}
-            <Menu.Menu position="right">
-              {hasPermission(profile, 'add_referraltoken') && (
-                <AddUserButton profile={profile} />
+    <Container fluid>
+      <AppBanner />
```
dankolbman, line 164: Does it make more sense to put it below the main navigation or above?

src/components/Header/AppBanner.js
```diff
@@ -0,0 +1,72 @@
...
+    const displayBanners = banners
+      .slice()
+      .sort((a, b) => a.startDate - b.startDate);
+
+    console.log('**** Displaying banners');
+    console.log(banners);
+
+    return (
+      displayBanners.length > 0 && (
+        <Container fluid className="app-banner" textAlign="center">
```
dankolbman, line 45: This container is very large and takes up a lot of real-estate on every page. Maybe scale it down to about half the size?

src/App.css
```diff
@@ -325,6 +325,22 @@ body {
     url(./assets/background-science.jpg);
 }
 
+.app-banner {
```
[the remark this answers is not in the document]
znatty22: Ah yeeea I do want to try to use the semantic components. The one thing I didn't figure out though with the segment is how to change the background color to something other than black which is what it gets set to with the inverted prop.

src/components/Header/AppBanner.js
znatty22: Ok, let me try moving it below the main nav bar and scaling it down. Do you think it should take up the full width of the page or no?

src/components/Header/Header.js
znatty22: Let me try both to see what it looks like

src/components/Header/AppBanner.js
dankolbman: Full page width is good so long as the content is within the container.

src/App.css
```diff
@@ -325,6 +325,18 @@ body {
     url(./assets/background-science.jpg);
 }
 
+.app-banner-bg {
+  background-color: rgb(195, 30, 124) !important;
```
dankolbman, line 329: Can you use an existing semantic color? It's important that we try to stay constrained to our color palette so that it doesn't become bloated.

src/components/Header/AppBanner.js
```diff
@@ -0,0 +1,77 @@
...
+    WARN: 'warning circle',
+    ERROR: 'times circle',
+  };
+
+  const isActive = (banner) => {
+    /* 
+    Banner is active if it is enabled && 
+    start date <= today && end date is after today 
+    */
+    if (!banner.startDate || !banner.endDate) {
```
dankolbman, line 21: Does it make more sense to consider a banner active if it is enabled and both its start and end dates are null? I would immagine most banners would not have a known lifetime when they are created with the exception of scheduled maintenance messages.

src/App.css
znatty22: Wellll, I tried the semantic pink but it was soo bright. Let me try purple or violet. Or maybe teal..

src/components/Header/AppBanner.js
znatty22: Hmm ok, I wasn't sure what kinds of messages we would be displaying. So a valid banner can be one without an end date

src/App.css
znatty22: Ohhh ok, so I think it was never correctly applying the semantic "pink" css before. That's why I thought it looked ugly. I figured it out. I just have to add "pink" to `className` prop on my Segment

src/admin/forms/BannerForm.js
```diff
@@ -0,0 +1,200 @@
+import React from 'react';
+import {withRouter} from 'react-router-dom';
```
dankolbman, line 2: HOCs aren't all that fashionable in React land any more as they've mostly been replaced by hooks: https://reactjs.org/docs/hooks-intro.html
I think in this case, the `useHistory` hook is what you're looking for.

XuTheBunny:
You may want to rebase master before merge using `git rebase master`.

XuTheBunny:
Maybe we can highlight (or add other marks) to indicate the current displaying banners.  
image

src/admin/views/BannersView.js
```diff
@@ -0,0 +1,109 @@
+import React from 'react';
+import {Helmet} from 'react-helmet';
+import {useHistory} from 'react-router-dom';
+import {useQuery, useMutation} from '@apollo/react-hooks';
```
XuTheBunny, line 4: You want to update this to import from @apollo/client - `import {useQuery, useMutation} from '@apollo/client';`

src/admin/views/NewBannerView.js
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import {Helmet} from 'react-helmet';
+import {useMutation} from '@apollo/react-hooks';
```
XuTheBunny, line 3: You want to update this to import from @apollo/client - import {useMutation} from '@apollo/client';

src/components/Header/AppBanner.js
```diff
@@ -0,0 +1,82 @@
+import React from 'react';
+import {Icon, Segment} from 'semantic-ui-react';
+import {useQuery} from '@apollo/react-hooks';
```
XuTheBunny, line 3: You want to update this to import from @apollo/client - import {useQuery} from '@apollo/client';

znatty22: Oh oops, good catch thanks @XuTheBunny!

dankolbman (approved):
# 📣

src/admin/common/bannerUtils.js
```diff
@@ -0,0 +1,35 @@
...
+};
+
+const activeDisplayBanners = banners => {
+  /*
+   * Display the 2 newest (by start date) banners
+   * */
+  return banners
+    .map(banner => banner.node)
+    .filter(banner => isActive(banner))
+    .slice(0, 2)
```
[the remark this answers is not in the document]
znatty22: Haha oops

PR #961, opened 2021-01-08
🐛 Fix footer backend version link

### LINKED ISSUE #953: API repo link at bottom of page is broken
## Relevant View or Screen

https://data-tracker.kidsfirstdrc.org

## Steps To Reproduce

Click on "v3.6" at bottom of page next to "Study Creator API"

Screen Shot 2020-12-23 at 11 37 28 PM

The linked page does not exist and you get a 404 on github.

## Expected Behavior

It should be working link instead of a 404.

XuTheBunny:
Screen Shot 2021-01-08 at 12 15 18 PM

- When  study-creator version number string is 'LOCAL' disable the link
- Version number string for study-creator in footer contains "v", it needs to be removed from the url.

Closes #953

### DIFF
--- a/src/components/Footer/Footer.js
+++ b/src/components/Footer/Footer.js
@@ -52,11 +52,13 @@ const Footer = () => {
               Study Creator API{' '}
               {status.version.split('-').length === 1 ? (
                 <a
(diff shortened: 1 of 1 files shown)


dankolbman (approved):
# 🍕

PR #968, opened 2021-01-19
🐛 Don't return exit code for copying artifacts

dankolbman:
`.test-result` only has contents from the docker compose command and doesn't contain anything after running `docker cp` so is not necessary.


XuTheBunny (approved):
🍥

PR #980, opened 2021-01-27
✨ Filter logs by job

### LINKED ISSUE #969: Filter job logs by job
## Feature Justification

As a developer, I would like to review logs for a particular task that I may be working on, rather than just the latest tasks.

Allow logs to be filtered by the job they belong to by selecting a log from the left pane:

Screenshot_2021-01-19 KF Data Tracker - Logs

dankolbman:
Allow users to filter logs by the job they belong to.

## Visual Changes

Screenshot_2021-02-08 KF Data Tracker - Logs

Closes #969

### DIFF
--- a/src/admin/components/JobsList/JobsList.js
+++ b/src/admin/components/JobsList/JobsList.js
@@ -1,33 +1,53 @@
 import React from 'react';
-import {Icon, List} from 'semantic-ui-react';
+import {Icon, Table} from 'semantic-ui-react';
 import TimeAgo from 'react-timeago';
 
-const JobsList = ({jobs}) => (
-  <List divided relaxed>
+const JobsList = ({jobs, onClick, active}) => (
+  <Table basic="very" celled compact="very">
+    <Table.Header>
+      <Table.Row>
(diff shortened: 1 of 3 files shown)


znatty22 (commented):
I like this feature but I'm wondering if the UI can be refactored a little. I think it takes a little too much mental effort to figure out what the components of each job list item mean. 

For example, I've guessed that the first line is the job name, clock icon means its a scheduled job etc, but its not immediately obvious. Also on first glance I missed which jobs were active vs inactive. 

Wondering if you could you turn that job list into a table instead so that these things are made more clear with the use of column names? I think there's enough horizontal real estate to do that and to display the filtered log list without issues. Also is it possible to reset the filter on the log list? I don't think there's a way to go back to viewing all logs without refreshing the page

dankolbman:
I swapped the menu for a table and added a remove filter button to reset to show all recent logs:
Screenshot_2021-02-08 KF Data Tracker - Logs

znatty22 (approved):
✅ Looks great!

PR #985, opened 2021-02-04
📈 Release charts

### LINKED ISSUE #972: Add cumulative releases graph to releases home view
## Feature Justification

It's encouraging to see the progression of releases occurring in the Data Tracker. We can display a simple graph of the cumulative sum of releases over time to indicate how active we have been in releasing data.

Add a graph to the home view with number of releases made over time.

dankolbman:
Add a statistics view for releases where we can summarize recent activity and performance of releases.

## Visual Changes

Screenshot_2021-02-05 KF Data Tracker - Releases

Closes #972

### DIFF
--- a/src/releases/views/StatsView.js
+++ b/src/releases/views/StatsView.js
@@ -1,12 +1,16 @@
 import React from 'react';
 import {Helmet} from 'react-helmet';
 import {useQuery} from '@apollo/client';
-import {GET_RELEASES} from '../queries';
+import {BASIC_RELEASES} from '../queries';
 import {Container, Header, Loader, Segment} from 'semantic-ui-react';
 import {ReleaseBarChart} from '../components/ReleaseBarChart';
+import {ReleaseActivity} from '../components/ReleaseActivity';
 
 const StatsView = ({match}) => {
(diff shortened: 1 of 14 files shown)


src/releases/views/StatsView.js
```diff
@@ -0,0 +1,45 @@
...
+  const releases =
+    releasesData && releasesData.allReleases.edges.map(({node}) => node);
+
+  return (
+    <Container as={Segment} basic vertical>
+      <Helmet>
+        <title>{`KF Data Tracker - Releases`}</title>
+      </Helmet>
+
+      <Header as="h1" className="noMargin">
```
znatty22, line 24: How about something like this to give more visual separation between the headers:

image

I removed the `noMargin` style on the first header and used `color=grey` style on both headers

znatty22 (commented):
Looks pretty cool! Had one suggestion for you

src/releases/views/StatsView.js
dankolbman: I agree that the margin shouldn't be removed here. Greying the text is a good practice for de-emphasizing without modifying text size, but I'm hesitant to do this because we don't follow this pattern with headers on other pages.

znatty22 (approved):
🎨

PR #995, opened 2021-02-26
🔥 Remove unused status fields

dankolbman:
The status query is now only used to display basic version info in the footer of the page. Our configuration page now uses a dynamic query that is generated and executed based on what fields exist on the status schema.

This fixes current issue of having a field moved on the status schema and will avoid any future issues where the settings and feature schema may change on the status model.


XuTheBunny (approved):
🍖

PR #1004, opened 2021-03-23
✨ Add data validation page for internal data review with validation report

### LINKED ISSUE #1034: Display loaders during transitory states for validation
## Feature Justification

See https://github.com/kids-first/kf-api-study-creator/issues/636

When the user clicks the `Cancel Validation` button, the validation run will be set to `canceling` state. Always display the canceling loader when the validation run is in the canceling state.

Similarly, when user clicks `Start Validation` button, the validation run will be set to `initializing` state and then `running` state once the validation begins executing. Display a loader with relevant state name during these states.

### LINKED ISSUE #1035: Display reason for validation failure
## Feature Justification

See https://github.com/kids-first/kf-api-study-creator/issues/637

When a validation run fails the user should see the reason for failure. The failure message will be accessible on the validation run object (see linked issue for exact name)

XuTheBunny:
Show data validation result on review list
Screen Shot 2021-04-20 at 3 42 10 AM

Show data validation result on review detail page
Screen Shot 2021-05-04 at 1 38 48 AM

Add data validation view with init, running, and completed stage
validation

Closes #1034
Closes #1035

### DIFF
--- a/src/documents/components/DataValidation/DataValidation.js
+++ b/src/documents/components/DataValidation/DataValidation.js
@@ -53,6 +53,9 @@ const DataValidation = ({
   const studyId = match.params.kfId;
 
   useEffect(() => {
+    if (validationRunState === 'running') {
+      startPolling(1000);
+    }
     if (validationRunState === 'failed') {
       stopPolling();
       setHasReport('failed');
@@ -208,7 +211,7 @@ const DataValidation = ({
                         },
                       })
                         .then(resp => {
-                          startPolling(10000);
+                          startPolling(1000);
                         })
                         .catch(err => {
                           setHasReport('not_started');
--- a/src/documents/components/DataValidation/DataValidation.js
+++ b/src/documents/components/DataValidation/DataValidation.js
@@ -15,6 +15,7 @@ import React, {useEffect, useState} from 'react';
 
 import {Amplitude} from '@amplitude/react-amplitude';
 import Markdown from 'react-markdown';
+import {dateCompare} from '../../utilities';
 import defaultAvatar from '../../../assets/defaultAvatar.png';
 import {longDate} from '../../../common/dateUtils';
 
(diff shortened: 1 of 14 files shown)


znatty22:
Oooo its coming alive 😄

src/documents/components/DataValidation/DataValidation.js
```diff
@@ -0,0 +1,244 @@
+import {
```
znatty22, line 1: @XuTheBunny Sorry for the long comment wasn't sure where to put all this. Also, I'm not sure if you were planning to implement some of this later once more of the validation was implemented on the backend. If so, I apologize for leaving feedback you might not need lol. But just in case:

**Validation Running**
User clicks the Click to Start Validation button, and then clicks the Run Validation button on the Validation view. Then the Running validation loader pops up. When the user navigates away from the Validation view while the validation is still running, the Data Review detail page still shows Click to Start Validation. Perhaps it should show a Running badge? Also, when the user clicks the Click to Start Validation button while the previous validation is running, the Validation view still shows the Run Validation button.

**Validation Failures**
In general it seems like the validation run failed state isn't taken into account yet. For example, the validation view doesn't update upon a validation run failure. It stays on the running validation loader. I think it would be useful to display the error message upon failure and then have a Re-run validation button somewhere.

**Data Review Detail View**
The data review detail view only shows the validation run states when the run is not started (Click To Start Validation) or when the run successfully completed (shows the result summary). Wondering if that area on the UI should also show the other states: failed, canceleld?

**Cancel Validation**
There doesn't seem to be a way to cancel a validation. Will this be added later?

**Re-run Validation**
There doesn't seem to be a way to re-run a failed validation. Will this be added later?

znatty22 (commented):
I tested this on a real file (using the study creator `do-validation` branch) and its cool to see it actually work! I left a bunch of comments inline but have one more I'll leave here. 

Could we add a download button the Validation view where the report is shown? The download button would allow a user to download the markdown report.

znatty22:
Hey @XuTheBunny could you add some kind of an icon for the validation run events? 

image

src/documents/views/ReviewDetailView.js
```diff
@@ -59,18 +59,37 @@ const ReviewDetailView = ({
...
     myProfile && hasPermission(myProfile, 'view_datareview');
+  const allowStartValidation =
+    myProfile && hasPermission(myProfile, 'add_validationrun');
 
   const [downloadFileMutation] = useMutation(FILE_DOWNLOAD_URL);
 
-  const {loading, data, error} = useQuery(DATA_REVIEW, {
-    variables: {
-      id: Buffer.from('DataReviewNode:' + reviewId).toString('base64'),
+  const {loading, data, error, startPolling, stopPolling} = useQuery(
```
znatty22, line 67: I noticed a lot of info being requested in the poll queries:

image

Some of the requested content seems unrelated to validation (e.g. study fields, dataReview.creator fields, dataReview.versions) and should probably be removed from the query. 

You may also want to revise the part of the query that fetches the validation runs. Right now the response will keep growing as more validation runs are created bc this query is fetching all of the validation runs on every poll. 

Maybe the poll requests should ask for the latest/last validation run? We might have to modify the validation runs query to make it easy to get the latest one.

src/documents/components/DataValidation/DataValidation.js
```diff
@@ -0,0 +1,388 @@
...
+                >
+                  Cancel Validation
+                </Button>
+              </Segment>
+            )}
+            {hasReport === 'failed' && (
+              <Segment placeholder>
+                <Header icon>
+                  <Icon name="x" />
+                  Failed to run validation report.
```
znatty22, line 313: There is an issue with the validation running loader displaying in a certain scenario:

1. User creates review, adds a file to review that will fail validation (e.g. spend.docx)
2. User clicks Click to Start Validation button
3. Validation starts running
4. Validation fails and validation view shows the Failed to run validation report message with Run Again button
5. User removes bad file from review, then adds good file (e.g. A GWO manifest) that will complete validation
6. User navigates to validation view and clicks Run Again button
7. The view doesn't update at all. The running loader with the cancel validation button is expected

znatty22 (changes requested):

znatty22 (commented):
Right now when a validation run fails, the validation view shows the message "Failed to run validation report". I think we should display the reason why a validation run failed as well. 

In order to do this, Data Tracker needs to access a failure message for the validation run. We could add a field on the validation run to store the reason for failure (something like `ValidationRun.failure_msg`). Thoughts?

src/documents/components/DataValidation/DataValidation.js
znatty22: I _think_ I'm seeing the same thing when I try to run a canceled run again by clicking Run Again button. The validation view doesn't change. I still see the Canceled message with the Run Again button.

znatty22:
@XuTheBunny I forgot to mention, it looks like the validation report markdown is still not rendering correctly. Idk if you got a chance to look into that yet

znatty22:
After discussion we realized some of the display issues are due to missing some necessary states in the backend. Additionally, we need to properly clear the validation results. I will implement these changes in https://github.com/kids-first/kf-api-study-creator/issues/636 and https://github.com/kids-first/kf-api-study-creator/issues/640. Then Xu will update this PR accordingly.

src/documents/components/ReviewDetail/ReviewDetail.js
```diff
@@ -338,6 +340,88 @@ const ReviewDetail = ({
...
+                        reviewNode.state,
+                      ) && canStartValidation ? (
+                        <>
+                          {validationRunState === 'running' ? (
+                            <Label
+                              as="a"
+                              color="blue"
+                              onClick={() => {
+                                history.push(
+                                  `/study/${match.params.kfId}/reviews/${reviewNode.kfId}/validation`,
```
shanewilson, line 396: I don't know much about Semantic UI - does it not allow you to pass `href` here?

znatty22 (approved):
🎉 ✅

PR #1011, opened 2021-04-01
🔥 Remove roles from user fragment

dankolbman:
Remove the soon to be deprecated `roles` from user queries. This field is not currently used in the UI or the backend.


XuTheBunny (approved):
🥩

PR #1030, opened 2021-05-05
✨ Update document types to match templates

### LINKED ISSUE #1016: Add/update document types to match templates
## Feature Justification

We need to update the document types to include the types accepted by the upcoming data validation feature. 

- Update the `fileTypeDetail` obj in `src/common/enum.js` to include the document types implemented in https://github.com/kids-first/kf-api-study-creator/pull/571
- Many of the templated document types (ones with required columns) are much more specific than the general document types. We should consider removing some of the general document types to avoid end user confusion. Perhaps we can keep the `Other` type as a catch-all and the `Sequencing Manifest` type since there is no template for it yet, but will be one in the near future.

znatty22:
Closes #1016 

## Visual Changes

image

### DIFF
--- a/src/common/enums.js
+++ b/src/common/enums.js
@@ -242,30 +242,6 @@ export const eventType = {
 
 // Store file type title, description and icon
 export const fileTypeDetail = {
-  SHM: {
-    icon: 'shipping',
-    title: 'Biospecimen Manifest',
-    requiredColumns: [],
-    description:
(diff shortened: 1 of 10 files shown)


znatty22:
Use different icon for Participant Observations (its the same as the demographics file)

znatty22 (commented):
👍 Looks good! Although I can't approve this PR since I opened it. We need an approval from @dankolbman

dankolbman (approved):
# 🎸

PR #1040, opened 2021-05-18
💄 Use Semantic header component replacing the default HTML header

XuTheBunny:
Instead of using `<h2>AAA</h2>` directly in frontend, use Semantic Header component as `<Header as='h2'>AAA</Header>`


dankolbman (approved):
# :dango:

PR #1057, opened 2021-06-08
🐛 Fix editing/linking/unlinking Cavatica projects error

XuTheBunny:
Screen Shot 2021-06-08 at 2 25 35 PM
Screen Shot 2021-06-08 at 2 25 43 PM


dankolbman (approved):
# :dollar:

PR #1044, opened 2021-05-18
✨ Organization switcher

dankolbman:
Quick feature to add a component for users to switch between their organizations. Only studies from the currently active organization will be displayed.

## Visual Changes

Screenshot_2021-05-18 KF Data Tracker - My Studies(2)

PR #1059, opened 2021-06-09
♻️ Attach organization to invites

dankolbman:
Uses the active organization when sending email invites from the invite modal or the add collaborator modal.

Depends on #1044


XuTheBunny (approved):
🍟

PR #1060, opened 2021-06-17
🐛 Fix allEvents query input event type

XuTheBunny:
Screen Shot 2021-06-17 at 2 33 14 AM
Screen Shot 2021-06-17 at 2 33 26 AM

Error:
`Variable \"eventType\" of type \"String\" used in position expecting type \"EventEventType\".`

Fixed by:
In `allEvents` query input change `$eventType: String` to `$eventType: EventEventType`

### DIFF
--- a/src/releases/components/StudyTable/StudyTable.js
+++ b/src/releases/components/StudyTable/StudyTable.js
@@ -50,7 +50,14 @@ const StudyTable = ({studies, selected, onChange}) => {
     name: node => node.name,
     kfId: node => <KfId kfId={node.kfId} key={node.kfId + 'kfId'} />,
     version: node => (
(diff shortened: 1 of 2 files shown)


dankolbman:
Was this a regression that got introduced maybe with the version upgrades or was it there already?

znatty22 (commented):
Just wondering when did this bug get introduced/how was stuff working before

dankolbman:
Looks like this change was introduced with an upgrade to the latest graphene: https://github.com/kids-first/kf-api-study-creator/commit/bf500248a313643d48cdabee2b82285375677ef9

dankolbman (approved):
👍

PR #1062, opened 2021-06-21
🐛 Fix study collaborator list display error

XuTheBunny:
Screen Shot 2021-06-21 at 1 54 03 AM

When a study has more than 1 collaborators, this error shows.
Screen Shot 2021-06-21 at 1 50 57 AM

Fixed by making a copy of user array before sorting method.
`users.sort()` -> `[...users].sort()`


dankolbman (approved):
# :hibiscus:

PR #1077, opened 2021-07-13
♻️ Replace static site orb with s3 orb

dankolbman:
The static site orb is trailing on the awscli setup script causing it to fail installation. This replaces the orb with circle maintained aws-s3.


XuTheBunny (approved):
🍞

PR #1116, opened 2021-09-08
✨ Add inline file mapping and editing with Flatfile

### LINKED ISSUE #1111: Inline file mapping and editing (Flatfile)
## Feature Justification

Users often do not want to or don't have the means to programmatically transform their data into the study's required templates. Right now there is no user friendly way or UI to help them do this. Instead most users undergo a very error prone process of copying and pasting rows or columns of data from one spreadsheet into another.

Additionally, if a user wants to make corrections or changes to a file in Data Tracker, they must download their file, make corrections locally and then re-upload their file as a new version.

It would be really nice if a user had a UI to help them do both of those things during the upload flow. We have found a product, Flatfile Portal, that can be integrated into Data Tracker as an import button to provide this functionality. Details are below:

## Flatfile Portal

Flatfile Portal is a button that when clicked opens up a UI for users to
upload a file, map their columns to template columns, make corrections to
the data, and review their file before completing the upload.

https://flatfile.com/product/portal/

## Demo + Code

Shane has already created a basic demo with just the Flatfile Portal button
using a couple of the NCI and Kids First templates

https://dewrangle-qf1f0gh3h-peykio.vercel.app/

Shane's code for the demo is:
https://github.com/kids-first/dewrangle/tree/main/src/components/Flatfile

### New Upload Flow

We need to provide a way for the user to "Try New Upload" via a toggle or
something on the UI. Then if the user decides to try the new upload experience
the flow would be something like this:

1. User clicks Upload Document (this would actually be the Flatfile button)
2. User is presented with Flatfile upload modal
3. User maps their file, fixes errors, and completes the upload in the modal
4. API creates new version
5. Resume same flows

### Backend

The Flatfile Portal button needs some configuration in order to work. The
configuration is essentially a list of template field definitions but
in the Flatfile format.

We may need to write a mutation that can convert our template into
their format. Or we may do this conversion on the frontend. Not sure yet.

XuTheBunny:
Screen Shot 2021-09-08 at 1 23 22 AM
Screen Shot 2021-09-09 at 2 08 27 AM
Screen Shot 2021-09-09 at 2 09 09 AM
Screen Shot 2021-09-09 at 2 10 02 AM
Screen Shot 2021-09-09 at 2 10 17 AM

Closes #1111

### DIFF
--- a/src/documents/components/FlatFile/FlatFile.js
+++ b/src/documents/components/FlatFile/FlatFile.js
@@ -1,0 +1,80 @@
+import React from 'react';
+import {Button, Popup} from 'semantic-ui-react';
+import {FlatfileButton} from '@flatfile/react';
+const {parse} = require('json2csv');
+
+/**
+ * FlatFile component for column mapping before document created
+ */
+
+const LICENSE_KEY = 'e118c06a-55ed-47be-90e4-26feb38c3934';
+
+const FlatFile = ({flatfileSettings, history}) => {
+  if (flatfileSettings.length > 0 && !flatfileSettings.startsWith('ERROR: ')) {
+    return (
+      <Button
+        as={FlatfileButton}
+        labelPosition="right"
+        icon="lightbulb"
+        content="Try New Upload"
+        color="violet"
+        licenseKey={LICENSE_KEY}
+        customer={{
+          companyId: 'CHOP',
+          companyName: 'CHOP',
+          email: 'test@example.com',
+          name: 'Admin',
+          userId: 'admin',
+        }}
+        settings={JSON.parse(flatfileSettings)}
+        onData={async results => {
+          const processedData = results.allData.map(row => {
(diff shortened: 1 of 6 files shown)


src/documents/components/FlatFile/FlatFile.js
```diff
@@ -0,0 +1,96 @@
...
+              } else {
+                formatedRow[key] = row[key];
+              }
+              return true;
+            });
+            return formatedRow;
+          });
+          console.log(processedData);
+          console.log(parse(processedData));
+          var file = new File([parse(processedData)], 'foo.csv', {
```
znatty22, line 47: You can grab the original file's name from https://flatfile.com/developers/sdk/classes/flatfileresults/#filename and use it here

znatty22:
Closing this since #1123 obsoletes this PR

PR #1117, opened 2021-09-10
✨ Better general file types

znatty22:
Bring back the old general file types and add additional useful general types:

image

When there are templates available for a study
image

PR #1118, opened 2021-09-10
🐛 Fix view only template missing function error on close

XuTheBunny:
Screen Shot 2021-09-13 at 11 08 35 AM


znatty22:
@XuTheBunny I think its better to have some useful general file types (beyond Other and dbGaP) that people can always select even if there are no templated types to choose from. See https://github.com/kids-first/kf-ui-data-tracker/pull/1117

PR #1134, opened 2021-10-04
♻️ Check if local stored organization belongs to user profile

### LINKED ISSUE #1125: Fix lifecycle of user organization storage
## Relevant View or Screen

image

## Steps To Reproduce

There is an issue with the lifecycle of a user's organization in local storage. We can produce this error in two ways:

### New user has no organization
1. New user signs up
2. New user tries to login
3. New user is greeted with a blank page due to not having an orgs

### Existing users have different organizations
UserA is a user that belongs to OrgA
UserB is a user that belongs to OrgB

1. UserA logs in and the current organization is set to OrgA
2. UserB logs in, they see a blank page

In this use case, it seems like OrgA is still cached in local storage but UserB is not a member of that organization so there is a runtime exception

## Solution
If the cached org doesn't exist, then query for the user's orgs and choose the first one. If the user has no organizations yet bc they are a new user, display some kind of page to them that says they haven't been added to an org yet and should contact KF support (similar to when there is an issue retrieving studies). Once user invites are working a user should always have an org, but until then, we can display this page so that new users aren't greeted with the white screen of death

XuTheBunny:
image

Closes #1125

### DIFF
--- a/src/components/Header/Header.js
+++ b/src/components/Header/Header.js
@@ -172,15 +172,21 @@ const Header = ({location}) => {
     setLoggedIn(true);
   }
 
-  const organizations = profile && profile.organizations;
+  // Check if an org is cached
   var currentOrg;
   try {
     currentOrg = JSON.parse(localStorage.getItem('currentOrganization'));
   } catch (e) {
     currentOrg = null;
   }
-  if (!currentOrg && loggedIn) {
-    currentOrg = organizations && organizations.edges[0].node;
+  // Check if stored current organization belongs to user profile
+  const organizations = profile && profile.organizations;
+  const hasOrgs = organizations && organizations.edges.length > 0;
+  const allowedOrgs = hasOrgs
+    ? organizations.edges.map(({node}) => node.id)
+    : [];
+  if (loggedIn && (!currentOrg || !allowedOrgs.includes(currentOrg.id))) {
+    currentOrg = hasOrgs && organizations.edges[0].node;
     localStorage.setItem('currentOrganization', JSON.stringify(currentOrg));
   }
 
@@ -193,7 +199,7 @@ const Header = ({location}) => {
 
   // Check if the org has changed on the backend and update localstorage
   if (
-    organizations &&
+    hasOrgs &&
     currentOrg &&
     JSON.stringify(
       organizations.edges.find(({node}) => node.id === currentOrg.id).node,


src/components/Header/Header.js
```diff
@@ -191,6 +191,16 @@ const Header = ({location}) => {
     },
   });
 
+  // Check if stored current organization belongs to user profile
```
znatty22, line 194: Line 182 is still going to cause issues when a user is logged in but doesn't have any organizations yet bc organizations.edges[0] is undefined. I'm trying to think how we want to refactor this logic .. 🤔

znatty22: I think you need something like this starting around line 175: 
```javascript
  // Check if an org is cached
  var currentOrg;
  try {
    currentOrg = JSON.parse(localStorage.getItem('currentOrganization'));
  } catch (e) {
    currentOrg = null;
  }
  // Check if stored current organization belongs to user profile
  const organizations = profile && profile.organizations;
  const hasOrgs = organizations && organizations.edges.length > 0;
  const allowedOrgs = hasOrgs
    ? organizations.edges.map(({node}) => node.id)
    : [];
  if (!currentOrg || !allowedOrgs.includes(currentOrg.id)) {
    currentOrg = hasOrgs && organizations.edges[0].node;
    console.log('Setting org ' + currentOrg);
    localStorage.setItem('currentOrganization', JSON.stringify(currentOrg));
  }
  const {refetch} = useQuery(ALL_STUDIES, {
    fetchPolicy: 'network-only',
    variables: {
      organization: currentOrg && currentOrg.id,
    },
  });

```
It's basically what you had before but removes that old block of code

src/components/Header/Header.js
```diff
@@ -172,15 +172,22 @@ const Header = ({location}) => {
...
-    currentOrg = organizations && organizations.edges[0].node;
+  // Check if stored current organization belongs to user profile
+  const organizations = profile && profile.organizations;
+  const hasOrgs = organizations && organizations.edges.length > 0;
+  const allowedOrgs = hasOrgs
+    ? organizations.edges.map(({node}) => node.id)
+    : [];
+  if (loggedIn && (!currentOrg || !allowedOrgs.includes(currentOrg.id))) {
+    currentOrg = hasOrgs && organizations.edges[0].node;
+    console.log('Setting org ' + currentOrg);
```
znatty22, line 190: 🔥 Sorry forgot to rm the log statement, could you please rm it?

znatty22 (approved):
🎃

PR #1153, opened 2021-12-07
♻️ Use API provided file name for template downloads

### LINKED ISSUE #1149: Use API provided file name for template downloads
Study creator API has a specific scheme for naming template packages. Data Tracker should use that filename instead of making one when the user downloads template(s). This can be parsed from the download response's content disposition header

XuTheBunny:
4 file name pattern:
Screen Shot 2021-12-07 at 1 48 30 PM

Closes #1149

### DIFF
--- a/src/components/TemplateList/TemplateList.js
+++ b/src/components/TemplateList/TemplateList.js
@@ -259,11 +259,17 @@ const TemplateList = ({
           let file = url;
           let headers = new Headers();
           headers.append('Authorization', bearerToken);
-          const fileName = `${studyKfid ? studyKfid : 'data'}_templates${
+          var fileName = `${studyKfid ? studyKfid : 'data'}_templates${
             format === 'zip' ? '.zip' : '.xlsx'
           }`;
           fetch(file, {headers})
-            .then(response => response.blob())
+            .then(response => {
+              fileName = response.headers
+                .get('Content-Disposition')
+                .split('filename*=UTF-8')[1]
+                .substring(2);
+              return response.blob();
+            })
             .then(blobby => {
               let objectUrl = window.URL.createObjectURL(blobby);
               anchor.href = objectUrl;
--- a/src/common/enums.js
+++ b/src/common/enums.js
@@ -263,6 +263,11 @@ export const eventType = {
     iconName: 'file excel outline',
     iconColor: 'red',
   },
+  TV_DOW: {
+    title: 'Template Version Downloaded',
+    iconName: 'download',
+    iconColor: 'green',
+  },
   DT_CRE: {
     title: 'Data Template Created',
     iconName: 'file excel outline',


src/components/TemplateList/TemplateList.js
```diff
@@ -259,13 +259,23 @@ const TemplateList = ({
...
+                .split('filename*=UTF-8')[1]
+                .substring(2);
+              response.blob();
+            })
             .then(blobby => {
-              let objectUrl = window.URL.createObjectURL(blobby);
+              var binaryData = [];
+              binaryData.push(blobby);
+              let objectUrl = window.URL.createObjectURL(
+                new Blob(binaryData, {type: 'application/zip'}),
```
znatty22, line 277: Will this always create a zip?

znatty22: Tested download of an excel workbook, and it doesn't open bc it thinks file is corrupted. Probably bc its downloaded as a zip

src/components/TemplateList/TemplateList.js
```diff
@@ -254,23 +254,12 @@ const TemplateList = ({
           const url = studyKfid
             ? `${KF_STUDY_API}/download/templates/${match.params.kfId}?file_format=${format}&template_versions=${tempList}`
             : `${KF_STUDY_API}/download/templates?file_format=${format}&template_versions=${tempList}`;
-          let anchor = document.createElement('a');
-          document.body.appendChild(anchor);
           let file = url;
           let headers = new Headers();
           headers.append('Authorization', bearerToken);
-          const fileName = `${studyKfid ? studyKfid : 'data'}_templates${
```
znatty22, line 262: Every time I try to download a template, I get a new window with `Not authorized toe download templates`. I think the logic you had before (everything you've removed) worked, and the only change needed is probably setting `anchor.download` to the fileName from the `Content-Disposition` header. Maybe something weird happened to this commit bc I don't see any `Content-Disposition` parsing anymore?

src/components/TemplateList/TemplateList.js
```diff
@@ -259,13 +259,27 @@ const TemplateList = ({
...
+                .substring(2);
+              response.blob();
+            })
             .then(blobby => {
-              let objectUrl = window.URL.createObjectURL(blobby);
+              var binaryData = [];
+              binaryData.push(blobby);
+              let objectUrl = window.URL.createObjectURL(
+                new Blob(binaryData, {
+                  type: fileName.endsWith('zip')
```
znatty22, line 278: Tried it and I'm getting the issue I had before, my machine isn't able to open any of the downloads bc it says the file is in an unsupported format. I see this for both zip and xlsx 

Btw, you should probable set this type directly from the response's `Content-Type` header

src/components/TemplateList/TemplateList.js
```diff
@@ -259,13 +259,27 @@ const TemplateList = ({
           let file = url;
           let headers = new Headers();
           headers.append('Authorization', bearerToken);
-          const fileName = `${studyKfid ? studyKfid : 'data'}_templates${
+          var fileName = `${studyKfid ? studyKfid : 'data'}_templates${
             format === 'zip' ? '.zip' : '.xlsx'
           }`;
           fetch(file, {headers})
-            .then(response => response.blob())
+            .then(response => {
```
znatty22, line 266: Is this bit returning both the fileName and the blob

znatty22 (approved):
✅

PR #1155, opened 2021-12-07
✨ Add filters for latest releases list

### LINKED ISSUE #1154: Add filters for latest releases list
Currently on production, we have hundreds of releases listed with 10 per page pagination.

It will be very convenient if user can filter through the releases.

Screen Shot 2021-12-07 at 12 41 46 PM

XuTheBunny:
Screen Shot 2021-12-09 at 1 06 19 AM
Screen Shot 2021-12-09 at 1 06 32 AM

Closes #1154

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -194,6 +194,10 @@ body {
     min-height: 350px !important;
 }
 
+.w-70 {
+  width: 70px !important;
+}
+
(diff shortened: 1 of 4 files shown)


znatty22 (approved):
📈

PR #1159, opened 2022-01-07
🐛 Fix event type filter from admin events page

### LINKED ISSUE #1158: Error out when clearing the event type filter from admin events page
Screen Shot 2022-01-06 at 11 51 54 AM
Screen Shot 2022-01-06 at 11 52 39 AM

XuTheBunny:
Passing empty string as event type cause graphql error.
Screen Shot 2022-01-06 at 11 54 11 AM

Fix by passing `null` value instead of empty string.
Screen Shot 2022-01-06 at 8 58 11 PM

Closes #1158


znatty22 (approved):
👍

